PR #507, opened 2014-09-25
Add missing cc-service-dashboards to main cf-properties.yml

drnic:
It already exists in warden template to set the secret, but full definition is missing from main properties list.

### DIFF
--- a/templates/cf-properties.yml
+++ b/templates/cf-properties.yml
@@ -277,3 +277,12 @@ properties:
         secret: (( merge ))
         authorities: cloud_controller.admin,scim.read
         authorized-grant-types: client_credentials
+      cc-service-dashboards:
+        secret: (( merge ))
+        override: true
+        authorized-grant-types: client_credentials
+        scope: cloud_controller.write,openid,cloud_controller.read,cloud_controller_service_permissions.read
+        authorities: clients.read,clients.write,clients.admin
+        access-token-validity: 1209600
+        refresh-token-validity: 1209600
+        autoapprove: true


gberche-orange:
@Amit-PivotalLabs its been two years since this feature is present in the code base. Are there reasons to leave the feature disabled by default ? 

Could the default be reconsidered, i.e. enable SSO dashboards by default as documented on http://docs.cloudfoundry.org/services/dashboard-sso.html#enabling-the-feature-in-cloudfoundry , instead of requiring Cf ops opt-in through explicit manifest tuning ?

thanks in advance,

Guillaume.

shalako:
+1

Amit-PivotalLabs:
I don't know personally know what this is, but it seems like it's probably fine.  The way this will work, it will require everyone to configure a secret for the `cc-service-dashboards` client.  Does that seem like a fair requirement to put on all users of the spiff templates?

This PR as is won't work, but we could:

* update the template in the correct place
* update all the test fixtures in the `spec/fixtures/${iaas}` directories
* update http://docs.cloudfoundry.org/services/dashboard-sso.html#enabling-the-feature-in-cloudfoundry if necessary
* update all the documentation about how to fill out the stubs, e.g. https://github.com/cloudfoundry/docs-deploying-cf/blob/1aecb9624aa4dc6ed5d00fbde47a766c4048aefa/aws/cf-stub.html.md.erb#L227-L245

Amit-PivotalLabs:
I'm going to update the template and spec fixtures so that they all do the same thing for all IaaSes (not have BOSH-Lite) be different.  I'm going to go with the standard configuration recommended in http://docs.cloudfoundry.org/services/dashboard-sso.html#enabling-the-feature-in-cloudfoundry.

Here's the story I'm working on.

PR #568, opened 2014-11-24
Adds default scopes for all UAA user tokens

ryanmoran:
When CF ships with notifications, users will need a couple new scopes to
be able to manage their notifications preferences. We are including
these scopes here so that all future users will have them.

[#83221726]


crhino:
Merged as of 3cc17b057966756af7511f1a26f49109e1158bd2

PR #586, opened 2014-12-29
Rename `loggregator` VMs to `doppler`

wfernandes:
The `loggregator` component was previously renamed to `doppler`, and shows up as
such in process names on the VM. However, we did not change the VM name at the
same time. This corrects that oversight to reduce ambiguity.

[#84731308]

Signed-off-by: Warren Fernandes <wfernandes@pivotal.io>


wfernandes:
Closing PR since it has diverged quite a bit from develop.

PR #587, opened 2014-12-31
Changed admin_pass/user to pass/user

longnguyen11288:
Changed admin_pass/user to pass/user to stay consistent with smoke test properties.


crhino:
Hi @longnguyen11288,

The acceptance tests actually require that you provide an admin user and password in order to succeed. In contrast to the smoke tests which do not require an admin user/password. Thus we feel uncomfortable changing the name of this property as it accurately reflects the input. Is there any reason you would like to make this change other than consistency in naming with the smoke tests?

Thanks,

@crhino && @jfmyers9

longnguyen11288:
Ahh I understand. Nope I was looking to make things consistent. It make sense.

PR #588, opened 2014-12-31
Allow max_in_flight for api jobs

adamstegman:


jpalermo:
@adamstegman We've made a change that should allow adding keys into the update hash for jobs

921787c3d0dc65e16c0416cbf5ed0a4c0db3ec91

PR #591, opened 2015-01-06
Add HAProxy to default security groups on bosh-lite

robdimsdale:
- for riak-cs, which needs to access the proxy and router as it uses http traffic.

[#79975532]

### DIFF
--- a/bosh-lite/cf-stub-spiff.yml
+++ b/bosh-lite/cf-stub-spiff.yml
@@ -33,6 +33,8 @@ properties:
       - name: services
         rules:
         - protocol: all
+          destination: 10.244.0.34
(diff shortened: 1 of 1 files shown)


zrob:
@robdimsdale can you make your membership of cloudfoundry public so the cla complaint goes away

also i think this makes sense as a separate group.  can you move it to a new security group rule called loadbalancer and add it to the list of default_running_security_groups?

thanks,
@zrob & @luan

robdimsdale:
I did make my membership public but the CLA hasn't updated to show that.

robdimsdale:
I see this is on release-candidate so I'm going to close this PR.

PR #593, opened 2015-01-08
Add merge to all _templates in cf-jobs.yml so they can be overridden

teancom:
I have been working on a cf release that uses a much reduced number of VMs (I'm attempting to get down to 4 + runners). My first approach was to not use cf-jobs.yml at all, just recreating the jobs myself. After being smacked down for that approach by @rkoster, I'm now attempting to use cf-jobs.yml, but "zero out" a bunch of the jobs, while redefining other templates (say, api_templates) to have more things it in (like gorouter and haproxy). However, unless the templates use '- <<: (( merge ))', they do not "pick up" my changes. As this will only affect those - like me - who are trying to override the existing configs, this should be a safe change to merge.

Thank you for your time and consideration.

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml
+++ b/spec/fixtures/aws/cf-manifest.yml
@@ -30,6 +30,7 @@ jobs:
         - 0.0.1.5
   resource_pool: router_z1
   templates:
+  - {}
   - name: haproxy
     release: cf
   - name: metron_agent
@@ -47,6 +48,7 @@ jobs:
       apps: cf1
   resource_pool: medium_z1
   templates:
+  - {}
   - name: nats
     release: cf
   - name: nats_stream_forwarder
@@ -66,6 +68,7 @@ jobs:
       zone: z2
   resource_pool: medium_z2
   templates:
+  - {}
   - name: nats
     release: cf
   - name: nats_stream_forwarder
@@ -87,6 +90,7 @@ jobs:
       zone: z1
   resource_pool: medium_z1
   templates:
+  - {}
   - name: etcd
     release: cf
   - name: etcd_metrics_server
@@ -107,6 +111,7 @@ jobs:
       zone: z2
   resource_pool: medium_z2
   templates:
+  - {}
   - name: etcd
     release: cf
   - name: etcd_metrics_server
@@ -124,6 +129,7 @@ jobs:
       zone: z1
   resource_pool: small_z1
   templates:
+  - {}
   - name: collector
     release: cf
   - name: metron_agent
@@ -139,6 +145,7 @@ jobs:
   persistent_disk: 102400
   resource_pool: medium_z1
   templates:
+  - {}
   - name: debian_nfs_server
     release: cf
   - name: metron_agent
@@ -154,6 +161,7 @@ jobs:
   persistent_disk: 4096
   resource_pool: medium_z1
   templates:
+  - {}
   - name: postgres
     release: cf
   - name: metron_agent
@@ -169,6 +177,7 @@ jobs:
       zone: z1
   resource_pool: medium_z1
   templates:
+  - {}
   - name: uaa
     release: cf
   - name: metron_agent
@@ -184,6 +193,7 @@ jobs:
       zone: z2
   resource_pool: medium_z2
   templates:
+  - {}
   - name: uaa
     release: cf
   - name: metron_agent
@@ -199,6 +209,7 @@ jobs:
       zone: z1
   resource_pool: medium_z1
   templates:
+  - {}
   - name: login
     release: cf
   - name: metron_agent
@@ -214,6 +225,7 @@ jobs:
       zone: z2
   resource_pool: medium_z2
   templates:
+  - {}
   - name: login
     release: cf
   - name: metron_agent
@@ -236,6 +248,7 @@ jobs:
       share: null
   resource_pool: large_z1
   templates:
+  - {}
   - name: cloud_controller_ng
     release: cf
   - name: metron_agent
@@ -260,6 +273,7 @@ jobs:
       share: null
   resource_pool: large_z2
   templates:
+  - {}
   - name: cloud_controller_ng
     release: cf
   - name: metron_agent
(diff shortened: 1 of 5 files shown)


teancom:
Ah, I see that it adds an extra '{}' which is throwing off the tests. Let me see if I can figure out how to fix that.

dieucao:
@teancom 
Could you help to get the new CLA signed by starkandwayne (assuming you are submitting this PR as part of starkandwayne) with a list of github usernames or github org specified that we can add to our CLA checking tool?  We recommend specifying a github org so that membership of the org can be managed by your team.  Team members would need to make membership of the org public.

dieucao:
@teancom I see that you just submitted on your CLA.  Thanks!

teancom:
I fixed the fixtures to add an empty list in all of the merged templates, which worked for *almost* all of the tests. There are two outstanding, one in warden, one in vsphere, that still fail - at least on my machine. But... I have no idea why. To my admittedly untrained eye, they look identical to the other jobs. I deleted and re-did the added line multiple times to rule out any sort of weird typo. If someone else could take a look at the failing tests (assuming they fail in the CI environment too), that'd be awesome.

luan:
Hi @teancom,

We were curious about why you decided to go with this strategy, we generally override the entire template section when we need to modify which templates are going to run in a given job.
An example of us doing that can be found [here](https://github.com/cloudfoundry/cf-release/blob/master/templates/cf-infrastructure-warden.yml#L289-L296).

We would prefer not adding an empty hash to every job if possible.

Thanks,
@luan & @DanLavine, CF Runtime Team

jbayer:
@dieucao i can confirm that @teancom has sent in a signed Cloud Foundry Foundation CLA

drnic:
Originally we wrote all new jobs; @rkoster posited that this wasn't the right approach. Any guidance you have on how we should construct massively adjusted/composed jobs would be appreciated.


Nic

rkoster:
I forgot the odd spiff behavior of merging an empty hash when there is nothing to merge. It could be that this will not work with bosh (should be tested) a usecase for this PR is for example wanting to collocate some sort of agent on different jobs. If you would use the approach of @luan this would result in a lot of maintenance since you would have to track all upstream job changes with regard to templates. With this PR you could just specify the collocated job in the meta section of your stub and be done with it.

teancom:
So, using this approach allows us to do this, in our own template:

```
api_templates:
 - name: gorouter
   release: (( meta.release.name ))
 - name: cloud_controller_ng
   release: (( meta.release.name ))
 - name: metron_agent
   release: (( meta.release.name ))
 - name: haproxy
   release: (( meta.release.name ))
 - name: nfs_mounter
   release: (( meta.release.name ))
```
if we want to consolidate a lot of templates onto one job, and reuse the existing api_z1/api_z2 jobs without any further configuration change (well, we'd also have to set the other, now redundant, instance counts to 0, which is trivial).

If I'm reading what you're saying right, you think the better approach would be to leave the template definition alone, and modify the job, like so?
```
jobs:
  - name: api_z1
    templates:
     - name: gorouter
     - name: cloud_controller_ng
     - name: metron_agent
     - name: haproxy
     - name: nfs_mounter
```
Unless I'm doing something wrong, that does not merge with the definition of the api_z1 job in cf-jobs.yml, it overwrites it completely. So we would end up having to do something like:

```
jobs:
  - name: api_z1
    templates:
     - name: gorouter
     - name: cloud_controller_ng
     - name: metron_agent
     - name: haproxy
     - name: nfs_mounter
    instances: 1
    resource_pool: large_z1
    persistent_disk: 0
    networks:
      - name: cf1
    properties:
      networks: (( meta.networks.z1 ))
      metron_agent:
        zone: z1
      nfs_server: (( meta.nfs_server ))
```
Which 1) leads to more duplication and 2) requires us to pull in definitions for meta.networks.z1 and meta.nfs_server into the current template, which is a pain and 3) requires us to keep our job definition "in sync" with upstream, tracking those changes (this is what @rkoster was referring to).

As I feel I'm still fairly n00b-ish when it comes to this, if I'm misunderstanding or misrepresenting how things work please feel free to smack me down, laugh at me, and kick sand in my face. Heaven knows @rkoster has done that before.

jfmyers9:
Hi @teancom,

Yes we are saying that we would prefer that you defined templates in your stub for each job individually such as:
```
jobs:
  - name: api_z1
    templates:
     - name: gorouter
     - name: cloud_controller_ng
     - name: metron_agent
     - name: haproxy
     - name: nfs_mounter
```
When you use the `./generate_deployment_manifest` tool specifying job templates in this manner should not override the other values defined under the job. Please feel free to edit the file in `./spec/fixtures/aws/cf-stub.yml` with the above changes and you should see how spiff will merge these changes into the final manifest. Please let us know if this works for you.

Thanks,

@jfmyers9 && @kdeshpan, CF Runtime Team

jfmyers9:
Hi @teancom,

With regards to staying in sync with upstream changes, how will adding these merge values to the template properties help you solve this problem? For example, if I want to collocate the gorouter on the cloud_controller, I may do something like:

```
meta:
  api_templates:
    - name: gorouter
      release: cf
```

This will add the gorouter template to the api_templates. However the gorouter job's templates are defined by the `meta.router_templates`. Thus you would need to add all components from `meta.router_templates` to `meta.api_templates` programmatically in order to ensure that you will stay in sync with future changes to `meta.router_templates`. I'm not sure of any way to achieve this through spiff currently, and it appears to me that you will still need to have some sort of manual check to ensure that you stay in sync with upstream changes in cf-release.

Thanks,

@jfmyers9 && @kdeshpan, CF Runtime Team

teancom:
To answer your second question first, we were thinking it would allow us to re-use most of the upstream job definition, by modifying the template definitions. We were actually hashing this out again this morning after our standup, and we think we might have an alternative solution that won't involve adding the empty arrays to each template (or any upstream cf change at all). Give me a bit to look into that to see if it works and I'll get back to you. Thank you very much for your quick and helpful responses!

Thanks,

David Bishop

luan:
Closing this per #598.

PR #595, opened 2015-01-13
Add bosh custom passwords

fbonelle:
We found that VMs deployed by Bosh for CloudFoundry were not sufficiently secure: the default password for root/vcap users (c1oudc0w) could not be changed easily.

What we want:
   - Possibility to change the password easily by resource_pool
   - Keep the use of Spiff to generate the final template

Our patch:
   - By default, the password is still c1oudc0w
   - Ability to set a new password overall

```
meta:
  ...
  default_env:
    # Default vcap & root password on deployed VMs (ie c1oudc0w)
    # Generated using mkpasswd -m sha-512
    bosh:
      password: $6$4gDD3aV0rdqlrKC$2axHCxGKIObs6tAmMTqYCspcdvQXh3JJcvWOY2WGb4SrdXtnCyNaWlrf3WEqvYR2MYizEGp3kMmbpwBC6jsHt0
    ...
```

   - Ability to set a password for each resource_pool

```
resource_pools:
    name: small_z1
    ...
    bosh:
      # Change default password:
      #   hashed password - Generate SHA hash using mkpasswd -m sha-512
      #   Generated hash for FT9KDyDoBqmeKQqEKoap
      password: $6$DVhgWycGMhiQYKsH$py04MGy0HVislvOLxtg/fx1RhycpbH7bSHBjxwZYC/KYg6gc59uN6LulsEk33WzggDQPBDjm0BM9iQXSzjITi1
    ...
```

### DIFF
--- a/templates/cf-deployment.yml
+++ b/templates/cf-deployment.yml
@@ -5,6 +5,12 @@ meta:
   # i.e. cf-tabasco
   environment: ~
 
+  default_env:
+    # Default vcap & root password on deployed VMs (ie c1oudc0w)
+    # Generated using mkpasswd -m sha-512
+    bosh:
+      password: $6$4gDD3aV0rdqlrKC$2axHCxGKIObs6tAmMTqYCspcdvQXh3JJcvWOY2WGb4SrdXtnCyNaWlrf3WEqvYR2MYizEGp3kMmbpwBC6jsHt0
+
   releases:
   - name: cf
     version: latest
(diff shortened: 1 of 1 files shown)


fbonelle:
@dieucao According to me, it should be backward compatible. By default, standard password is kept. This patch only allow to specify the tag bosh in resource_pools spec and being correctly merged by spiff.

jpalermo:
Hi @fbonelle 

The VMs deployed by BOSH are only accessible if you have the private ssh key. I believe the usernames and passwords are only there because the OS requires some sort of username and password.

If you remove your ssh keys, there should be no way to ssh into one of the VMs. Are you seeing different behavior?

fbonelle:
Hi @jpalermo 

What we have seen:
   * By default, there is no ssh keys on the VMs deployed by Bosh ?
   * For our deployments, we take VSphere or VCloud stemcell. With these stemcells, there is a default password that can be customized (threw bosh template). 

I didn't really found a documentation or tutorial explaining how pushing ssh keys to deployed VMs with bosh yaml template file.

Thanks

jpalermo:
@fbonelle 
We confirmed what you are seeing. The behavior is different for VSphere and AWS. We'll see about getting this merged in.

Thanks for the follow up.

PR #615, opened 2015-02-11
Add networks properties to postgres and nfs

longnguyen11288:
Added networks properties so it can be used in other colocated jobs such as consul


longnguyen11288:
Made PR to wrong branch

PR #616, opened 2015-02-11
Added networks properties for postgres and nfs

longnguyen11288:
Added properties to postgres and nfs so colocated jobs can find their own ip. eg consul


zrob:
Hey @longnguyen11288 

This seems like a simple, reasonable change.  However, those values are not accessible without a corresponding update to the spec files for those jobs.

Thanks,
@zrob & @zaksoup

longnguyen11288:
Do you want me to add this to postgres even though they aren't using it? I'm adding it as a standard so colocated jobs can use them. I use it in my consul-cf integration.

PR #617, opened 2015-02-11
Remove unused properties from uaa job spec.

xian:
Unused in templates. Below is the script we used to find unused keys; might be useful as a linter?

```ruby
#!/bin/env ruby

require 'yaml'

job_dir = '/Users/pivotal/workspace/cf-release/jobs/uaa'
spec = YAML.load_file("#{job_dir}/spec")
properties = spec['properties'].keys

erb_files = Dir["#{job_dir}/templates/*.erb"]
contents = {}
erb_files.each { |f| contents[File.basename(f)] = File.read(f) }

used = {}
properties.each do |property|
  contents.each do |file, str|
    if str.include?(property)
      (used[property] ||= []) << file
    end
  end
end

puts "Used properties:"
used.each { |property, files| puts "* #{property} used by: #{files.sort.join(", ")}"}

puts
puts "Unused properties:"
(properties - used.keys).each { |property| puts "* #{property}" }
```

### DIFF
--- a/jobs/uaa/spec
+++ b/jobs/uaa/spec
@@ -31,10 +31,6 @@ properties:
     description: "Number of seconds in which lockoutAfterFailures failures must occur in order for account to be locked"
   uaa.authentication.policy.lockoutPeriodSeconds:
     description: "Number of seconds to lock out an account when lockoutAfterFailures failures is exceeded"
-  uaa.batch.password:
-    description: "Deprecated"
-  uaa.batch.username:
-    description: "Deprecated"
   uaa.catalina_opts:
     description:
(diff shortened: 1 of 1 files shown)


drnic:
Perhaps instead of removing from spec they need adding into templates?

xian:
Pretty sure at least the `ccdb` entries are irrelevant.

zrob:
Hey Christian,

I'll send this over to UAA to take a look.  Could you make your cloudfoundry org membership public so that cfdreddbot doesn't complain about CLAs

Thanks,
Zach & Zak

fhanik:
You have our approval, some of them even say deprecated, and have probably been like that for a very long time.

PR #622, opened 2015-02-20
Set default to 64000 max conn

longnguyen11288:
This is because even though it set in global it limit is only 2000. Refer to http://stackoverflow.com/questions/8750518/difference-between-global-maxconn-and-server-maxconn-haproxy for more info.

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -9,6 +9,7 @@ global
 defaults
     log global
     timeout connect 30000ms
+    maxconn 64000
     <% if p("request_timeout_in_seconds").to_i > 0 %>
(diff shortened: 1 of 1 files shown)


dieucao:
This change was merged.  Not sure why it's not showing up here though.
https://github.com/cloudfoundry/cf-release/commit/641a6b8358607e58cafecf0fb9d79d44ea1308f8

ematpl:
Thanks, @longnguyen11288! In the future, please make pull requests against the develop branch of cf-release, as is mentioned in the [contributing guidelines](https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md#general-workflow). They will then update correctly on GitHub when the Runtime team merges them into that branch. 

Thanks again,
@ematpl

longnguyen11288:
Ah. Damn. I always forget to make it to develop. I will make sure to do it
in future

PR #627, opened 2015-03-03
Adding X-Forwarded-Proto header to HTTP check.

amhuber:
Modifying the HTTP check in monit for UAA and Login to include the X-Forwarded-Proto header.  This ensures that the check will function when uaa.require_https = true and login.protocol = https in the configuration.  This change was discussed and approved in https://github.com/cloudfoundry/uaa/issues/127.


dieucao:
@amhuber This has been merged.  It should be available in the next cf-release.
https://github.com/cloudfoundry/cf-release/commit/647558bf5602749ec80dd89f700b60ba7e2125bc

PR #644, opened 2015-03-23
HAProxy: Always set X-Forwarded-For header

tsl0922:
From the [HAProxy 1.2 doc](http://www.haproxy.org/download/1.2/doc/architecture.txt):
 
- if the application needs to log the original client's IP, use the
   "forwardfor" option which will add an "X-Forwarded-For" header with the
   original client's IP address. You must also use "httpclose" to ensure
   that you will rewrite every requests and not only the first one of each
   session :

        option httpclose
        option forwardfor

This also works for CF's packaged HAProxy(1.5).

### DIFF
--- a/jobs/haproxy/templates/haproxy.config.erb
+++ b/jobs/haproxy/templates/haproxy.config.erb
@@ -20,6 +20,7 @@ frontend http-in
     mode http
     bind :80
     option httplog
+    option httpclose
     option forwardfor
     reqadd X-Forwarded-Proto:\ http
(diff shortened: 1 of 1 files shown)


tsl0922:
This problem exists on my CF instance(v172) , with an outdated version of HAProxy(1.5-dev19).

It seems this option is not needed for the latest version of HAProxy(1.5.11), I've tested the HAProxy of the current master branch(1.5.10), this option is not needed too.

PR #646, opened 2015-03-25
Expose signRequest and signMetaData via the login job

johnmcteague:
signRequest and signMetaData are config options of login.saml (see https://github.com/cloudfoundry/uaa/blob/4ae9109107102bcf0b3849f5aa964abde4629548/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml#L94 and https://github.com/cloudfoundry/uaa/blob/4ae9109107102bcf0b3849f5aa964abde4629548/uaa/src/main/webapp/WEB-INF/spring/saml-providers.xml#L86)

However there is no way to configure these via login job.

This PR adds two new options to the login spec file and maps them through to the login.yml  template.

### DIFF
--- a/jobs/login/spec
+++ b/jobs/login/spec
@@ -117,6 +117,12 @@ properties:
     description: "Deprecated: Use login.saml.providers list objects"
   login.entity_id:
     description: "Deprecated: Use login.saml.entityid"
+  login.saml.signMetaData:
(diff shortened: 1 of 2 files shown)


jfmyers9:
Merged as of aacc01eea86b4670b782a3cc55bcebbb7bf01af9

PR #651, opened 2015-04-05
Updated to have the latest generation of AWS instances.

mb002270:
I migrated m1.XX to m3.XX, however, m1.small is not offered as an m3 instance type.  Migrated m1.small to m3.medium.  

Also, m1.small and isn't available in all AZs (e.g. us-east-1e), so if you don't accept my full pull request, please at least update the m1.small to m3.medium to prevent deployment errors.

### DIFF
--- a/templates/cf-infrastructure-aws.yml
+++ b/templates/cf-infrastructure-aws.yml
@@ -50,7 +50,7 @@ properties:
 
 compilation:
   cloud_properties:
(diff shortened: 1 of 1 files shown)


mb002270:
I fixed the broken build (sorry about that).  I have pushed the changes to my fork.  I hope you can see them (newish to github).  Let me know if I need to do something else.  Thanks.

PR #652, opened 2015-04-05
Add timeout related options for acceptance tests.

simonjohansson:
I got the default values from

https://github.com/cloudfoundry/cf-acceptance-tests/blob/fa899ceaa17c4f1fde50218eac1a5cad4ee17a28/apps/init_test.go#L14
and
https://github.com/cloudfoundry/cf-acceptance-tests/blob/e097be328c334647b812b5f49ebbab40899d714f/services/init_test.go#L14

### DIFF
--- a/jobs/acceptance-tests/spec
+++ b/jobs/acceptance-tests/spec
@@ -51,3 +51,15 @@ properties:
   acceptance_tests.include_v3:
     default: false
     description: Flag to include the v3 API test suite.
+  acceptance_tests.default_timeout:
+    default: 30
+    description: Seconds to wait before cf cli interactions will time out.
+  acceptance_tests.cf_push_timeout:
+    default: 120
+    description: Seconds to wait before cf cli interactions against an app will time out.
+  acceptance_tests.long_curl_timeout:
+    default: 120
+    description: Seconds to wait before a curl against an app will time out.
+  acceptance_tests.broker_start_timeout:
+    default: 300
+    description: Seconds to wait before cf cli interations in the context of service brokers will time out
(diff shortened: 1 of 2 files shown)


DanLavine:
hi @simonjohansson,

It looks like the manifest requires integer values for these default timeouts. Setup for the defaults can be found on the acceptance-tests [readme](https://github.com/cloudfoundry/cf-acceptance-tests/blob/master/README.md). Configuration is something like:

```yml
  "default_timeout": 45,
  "cf_push_timeout": 180,
  "long_curl_timeout": 180,
  "broker_start_timeout": 300
```

After theses changes we can pull in the PR.

Thanks,
@DanLavine && @utako, CF Runtime Team

simonjohansson:
Good catch.

zrob:
Hi @simonjohansson,

It looks like we're already configuring the default timeouts in the CATS codebase, and we would like for that to be the source of truth for defaults. Could you please remove the defaults from the spec file to prevent any overlap? We'll be happy to pull in the PR after that.

Thanks,
@zrob && @utako, CF Runtime Team

simonjohansson:
@zrob The reason for the defaults in the spec is because it seems a bit silly to 

```
if default_timeout
  default_timeout: ...
end
if cf_push_timeout
  cf_push_timeout: ...
end
...
...
```
in the template.

But sure, I get your point.

simonjohansson:
Im afraid I did not have time to look at this today, and starting from tomorrow I will be on holiday for 2 weeks. Feel free to close and Ill reopen when I get back if someone else hasn't added support for this.

zaksoup:
Feel free to reopen when you return

Zak

PR #653, opened 2015-04-06
Updating AWS Infrastructure to be current generation of instance types

mb002270:
I updated the aws infrastructure yml file to have the latest generations of instance types.

This saves users money since the m1.X instance type and the c1.X instance types cost more money per hour than their current generation counterparts. 

The migration was as follows:
m1.small -> m3.medium (there is no m3.small offering)
m1.medium -> m3.medium
c1.medium -> c3.large (there is no c3.medium offering)

Updated unit tests also.

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml.erb
+++ b/spec/fixtures/aws/cf-manifest.yml.erb
@@ -1,7 +1,7 @@
 compilation:
   cloud_properties:
     availability_zone: zone_1
-    instance_type: c1.medium
+    instance_type: c3.medium
   network: cf1
   reuse_compilation_vms: true
   workers: 6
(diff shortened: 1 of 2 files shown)


dieucao:
@mb002270 We are actually working on updating the aws defaults in cf-release to new instance types in a separate story.
Initial story. https://www.pivotaltracker.com/story/show/89937452
Follow up story to adjust disk sizes. https://www.pivotaltracker.com/story/show/91780082
Follow up story to investigate better defaults on disk sizes https://www.pivotaltracker.com/story/show/91780134

Also, in general we found that m3.medium aren't actually equivalent to m1.medium.  We did not choose m3.medium in any scenario.

PR #655, opened 2015-04-09
Updated AWS instance types for current generation

geofffranks:
Updated the instance types for AWS services to use current generation equivilents (lower cost, higher performance IO than previous).

There is no m3.small (for the m1.small upgrade path), and m3.medium only has 4GB of local storage, which seemed like it might not be enough, so I went with m3.large across the board for those.

### DIFF
--- a/templates/cf-infrastructure-aws.yml
+++ b/templates/cf-infrastructure-aws.yml
@@ -50,7 +50,7 @@ properties:
 
 compilation:
   cloud_properties:
-    instance_type: c1.medium
+    instance_type: c3.large
     availability_zone: (( meta.zones.z1 ))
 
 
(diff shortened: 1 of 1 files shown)


dieucao:
@geofffranks 
We are actually working on updating the aws defaults in cf-release to new instance types in a separate story.
Initial story. https://www.pivotaltracker.com/story/show/89937452
Follow up story to adjust disk sizes. https://www.pivotaltracker.com/story/show/91780082
Follow up story to investigate better defaults on disk sizes https://www.pivotaltracker.com/story/show/91780134

This initial story will be part of cf-release 206.

geofffranks:
Even better, thanks!

PR #657, opened 2015-04-10
define separate fog aws key

leizhu:
To deploy cf on Tokyo region in AWS, we need two AWS keys. one is for AWS resource (e.g. VM, subnet..) . And we also need another key to access buildpack. So we defined a new property fog_aws_key for AWS key to access buildpack.

### DIFF
--- a/templates/cf-infrastructure-aws.yml
+++ b/templates/cf-infrastructure-aws.yml
@@ -5,8 +5,8 @@ meta:
 
   fog_config:
     provider: AWS
-    aws_access_key_id: (( properties.template_only.aws.access_key_id ))
-    aws_secret_access_key: (( properties.template_only.aws.secret_access_key ))
+    aws_access_key_id: (( properties.fog_aws_key.aws_access_key_id ))
+    aws_secret_access_key: (( properties.fog_aws_key.aws_secret_access_key ))
     region: us-east-1
 
   stemcell:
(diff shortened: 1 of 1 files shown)


dieucao:
@leizhu Can you make your membership of the CloudFoundry org public?

leizhu:
@dieucao , now I have made my membership public

jfmyers9:
Hi @leizhu,

I'm not sure that we completely understand the intent from the commit. It appears to just shift references around and add a level of indirection. Is there some matching code we're not seeing here? Can you provide us with an example of how you will use these new properties in order to help us better understand this change?

Thanks,

@jfmyers9 && @sergueif, CF Runtime Team

leizhu:
@jfmyers9  @jim80net 

You can take a look at the file https://github.com/pivotal-cloudops/birdnest-aws/blob/master/stubs/cf-aws-stub.yml

After we added a new reference to the properties, e.g. **aws_access_key_id**. Stub file can override them and set different values to **fog_aws_key. aws_access_key_id** and **template_only.aws. aws_access_key_id**.  In our deployment environment, we need to set different values for them. As we said in the pull request, we need to set different values to access AWS VM/Subnet resource and buildpacks.

zaksoup:
Hey @leizhu, I had a conversation with @jim80net. We'll be following up about a better way to set your stubs up to avoid having to make this change in the cf-release templates.

PR #661, opened 2015-04-17
Upgrading nodejs buildpack to v1.3.0

cf-buildpacks-eng:
Upgrading to latest stable version of nodejs buildpack
 --Buildpacks Team


jtarchie:
Closing because @flavorjones would like to follow up on vcap-dev before official merge.

PR #662, opened 2015-04-22
Upgrading ruby buildpack to v1.3.1

cf-buildpacks-eng:
Upgrading to latest stable version of ruby buildpack
 --Buildpacks Team


flavorjones:
Just a note that this buildpack addresses CVE-2015-1855

PR #667, opened 2015-04-28
Disable SQL connection timeouts 

cf-identity:
for the sake of migrations completing successfully

https://www.pivotaltracker.com/story/show/93459408
[#93459408]


dieucao:
@cf-identity Can you make your membership of the cloudfoundry org public so that dreddbot doesn't flag your PRs?  Also, please add an avatar =D

cf-identity:
@dieucao I made the pivotal.io email address public, that should suffice for the CLA check right?
Also, added a public domain, free to reuse avatar for our Born Again Identity

PR #670, opened 2015-05-05
Be able to specify timeouts for acceptance tests without defaults in the spec.

simonjohansson:
My previous PR (https://github.com/cloudfoundry/cf-release/pull/652) got rejected

"It looks like we're already configuring the default timeouts in the CATS codebase, and we would like for that to be the source of truth for defaults. Could you please remove the defaults from the spec file to prevent any overlap? We'll be happy to pull in the PR after that."
@zrob

It seems the similar changes ended up in develop anyways
https://github.com/cloudfoundry/cf-release/commit/b6c1f33771213ded1cf7c982f5f6fafb3d900197

Anyways, here is a new PR that addresses the issue that got raised.

### DIFF
--- a/jobs/acceptance-tests/spec
+++ b/jobs/acceptance-tests/spec
@@ -55,14 +55,10 @@ properties:
     default: false
     description: Services tests push their apps using diego if enabled
   acceptance_tests.default_timeout:
-    default: 30
     description: Default Timeout
   acceptance_tests.cf_push_timeout:
-    default: 120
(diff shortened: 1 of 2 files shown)


utako:
Hi @simonjohansson,

Thanks for updating the changes since our last conversation. Unfortunately our pipeline is blocked right now, but we will pull in your changes once it becomes clear. 

@DanLavine && @utako, CF Runtime Team

PR #671, opened 2015-05-07
Upgrading nodejs buildpack to v1.3.1

cf-buildpacks-eng:
Upgrading to latest stable version of nodejs buildpack
 --Buildpacks Team


flavorjones:
Please note that this is the "skinny" buildpack.

PR #672, opened 2015-05-08
Add missing description in jobs

mmc2004jp:
This PR is to add missing description for the properties in 'spec' files.

### DIFF
--- a/jobs/cloud_controller_clock/spec
+++ b/jobs/cloud_controller_clock/spec
@@ -27,19 +27,19 @@ properties:
 
   name:
     default: "vcap"
-    description:
+    description: "Name showing at /info endpoint"
   build:
     default: "2222"
-    description:
+    description: "build number of cloud foundry"
   version:
     default: "2"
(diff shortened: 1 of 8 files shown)


fhanik:
Thanks for the PR. I believe this is a good start. The descriptions missing is definitely bad, but I think when there is a description it could have a full description, such as what the property means, what type of values it may have and how it is used by the job. For example  jobs/cloud_controller_clock/spec and the property login.url, I'm not sure what the description means that the login job will register this URL with the router, and what impact this actually has on the cloud_controller_clock job. The fact that the property is defined in here, implies that the cloud_controller_clock uses this URL to reach the login server, rather than router registration.

zaksoup:
@fhanik Those are good points about descriptions for items that may not be perfectly obvious (uaa.url in the cc). I'm going to create a chore for CF-release to do this stuff starting on the work done here. Closing this for now. Thanks @mmc2004jp for getting this kicked off!

PR #675, opened 2015-05-08
Upgrading ruby buildpack to v1.4.1

cf-buildpacks-eng:
Upgrading to latest stable version of ruby buildpack
 --Buildpacks Team


mhoran:
Please note that this is the "skinny" buildpack.

PR #680, opened 2015-05-14
staticfile to be tested before nodejs/ruby buildpacks

drnic:
Many staticfile apps are using nodejs or ruby toolchains to construct their assets. This can trick the `nodejs_buildpack` or `ruby_buildpack` to try to deploy the app. I suggest placing `staticfile_buildpack` earlier in the list.

### DIFF
--- a/templates/cf-properties.yml
+++ b/templates/cf-properties.yml
@@ -144,6 +144,8 @@ properties:
     user_buildpacks: (( merge || [] ))
     system_buildpacks: (( merge || default_buildpacks ))
     default_buildpacks:
+      - name: staticfile_buildpack
+        package: buildpack_staticfile
       - name: java_buildpack
         package: buildpack_java
       - name: ruby_buildpack
@@ -156,8 +158,6 @@ properties:
         package: buildpack_python
       - name: php_buildpack
         package: buildpack_php
-      - name: staticfile_buildpack
-        package: buildpack_staticfile
       - name: binary_buildpack
         package: buildpack_binary
 


flavorjones:
This seems reasonable, as the `staticfile-buildpack` will only be activated by the presence of a `Staticfile` file. No objections.

dieucao:
@drnic @flavorjones 
Thoughts on moving this to after the java buildpack?
In the auto detect order I think it makes sense to auto detect java first if that doesn't cause any conflicts.

drnic:
Why does it make sense?

dieucao:
@drnic For auto detect, my feeling is that since the highest percentage of apps are java apps on first push it would be desirable to try the java buildpack first.  Does that sound reasonable?

drnic:
Are there any build systems for static apps that are written in Java? That is, if you used a Java app or jar to create your html/CSS/JavaScript then would the Java build pack accidentally think you're running a Java app? Where as you know it's a staticfile app because of Staticfile?




Can you point me to docs on how Java bin/detect behaves?




I know how staticfile build pack behaves and it would never accidentally deploy a Java app.

dieucao:
@drnic 
On the java-buildpack [README](https://github.com/cloudfoundry/java-buildpack/blob/master/README.md) if you click into each of the standard containers and standard frameworks you can see how they each do detection.

@nebhale @cgfrost
Do you have any concerns about staticfile-buildpack being first in the autodetect order.

nebhale:
@dieucao @drnic Since Java applications are built locally and pushed in their "runnable" form, placing the `staticfile_buildpack` ahead of the Java buildpack should be safe.  It's possible that someone could place a `Staticfile` in the root of their Java application, but I'd consider that an aberration that should be handled by specifying the Java buildpack.

dieucao:
This was merged and should be available in the next cf-release.
Not sure why it doesn't show up here as merged but the commit is here:
https://github.com/cloudfoundry/cf-release/commit/ff7a30552163d0127b278f75b95e86698f542e6c

PR #687, opened 2015-05-26
Readme clarifications

jvshahid:
Clarify that the security group name should be used instead of the security group id

### DIFF
--- a/example_manifests/README.md
+++ b/example_manifests/README.md
@@ -27,7 +27,7 @@ it doesn't include features such as high availability and security.
 - Click "Save"
 
 - Replace REPLACE_WITH_DIRECTOR_ID in the example manifest with the bosh director id (return by running "bosh status --uuid")
-- Replace REPLACE_WITH_BOSH_SECURITY_GROUP in the example manifest with the security group created for microbosh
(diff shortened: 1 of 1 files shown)


flavorjones:
Hi there!

We use Pivotal Tracker to provide visibility into what our team is working on. A story for this issue has been automatically created.

The current status is as follows:

* [ ] [#99824324](https://www.pivotaltracker.com/story/show/99824324) Readme clarifications


This comment, as well as the labels on the issue, will be automatically updated as the status in Tracker changes.

Amit-PivotalLabs:
Hi @jvshahid,

It would be preferable to not have the BOSH error message in the README, as their text output might change over time and this might become confusing to someone using this README and seeing a different error message.

Hopefully this README can go away entirely at some point and be replaced with cloudformation templates which will be much easier for anyone to use (and testable), and documentation on `bosh.io` or `docs.cloudfoundry.org` for deploying BOSH and/or cloud foundry, so there's a single source of truth that isn't duplicated here.

That said, if you'd like to submit a PR against `develop` (not `master`) with the README improvements, minus the BOSH output, we'd be happy to merge them in.  I'll close out this PR for now.

Thanks,
Amit

Issue #690, opened 2015-05-28
Manifest templates missing metron agent deployment property


mjseid:
Using v210 and the generate manifest script, the created template gives the following error:

 Started preparing configuration > Binding configuration. Failed: Error filling in template `metron_agent.json.erb' for `ha_proxy_z1/0' (line 5: Can't find property `["metron_agent.deployment"]') (00:00:00)

Error 100: Error filling in template `metron_agent.json.erb' for `ha_proxy_z1/0' (line 5: Can't find property `["metron_agent.deployment"]')

See related issue:
https://github.com/cloudfoundry/bosh-lite/issues/265

djvdorp:
We're running into the exact same issue as described by @cdavisafc (via @mjseid, found above).

jpalermo:
We're not seeing this problem.

`metron_agent.deployment` should be getting filled in by `cf-release/templates/cf-lamb.yml`

Could you provide the following to help us troubleshoot?

OS, spiff version, and which infrastructure you are deploying to.

Thanks.

mjseid:
Sure.  I do have the cf-lamb.yml you referenced, and it does contain some metron agent deployment specs as the last 2 lines of the file.
OS: ubuntu
Spiff: 1.0.3
IaaS: Vsphere

jpalermo:
Hmm, and in the deployment manifest that gets generated. Do you have a `properties.metron_agent` section at the top level?

mjseid:
there is a metron_agent section that gets created under top level properties:
  metron_agent:
    deployment: null

But then each job only gets the zone property and not a deployment property Ex:
jobs:
- instances: 0
  name: consul_z1
  networks:
  - name: cf1
    static_ips: []
  persistent_disk: 1024
  properties:
    consul:
      agent:
        mode: server
    metron_agent:
      zone: z1
  resource_pool: medium_z1
  templates:
  - name: consul_agent
    release: cf
  - name: metron_agent
    release: cf
  update:
    max_in_flight: 1
    serial: true

jpalermo:
Yes, each job will not get the property, it is only expected to live at the top level.

BOSH will apply any top level properties to every job, then the job specific properties will overwrite those.

Do you have the `meta.environment` property configured in your stub file?

mjseid:
Nope, I didn't have that setting....guess things have worked fine til now without it. Upon setting it, and re-generating a manifest the resulting yml is identical to before though.  Here's how my stub looks now:
```
---
[...]

PR #696, opened 2015-06-03
Fix for templates missing metron agent deployment property

gberche-orange:
Fix for https://github.com/cloudfoundry/cf-release/issues/690

Added missing metron_agent.deployment properties, including in loggregator (bumping commit id)


gberche-orange:
Sorry, I was caught by a lack of "git submodule update" which had left cf-release/templates/cf-lamb.yml outdated.

Sorry for the noise and extra work involved reviewing this PR.

PR #708, opened 2015-06-19
Removed `--force` - line38

kennetham:
Please double confirm if `--force` is required. The git that I am running on suggested that there isn't such a command.


wendorf:
Hi @kennetham,

It looks like you are using an older version of git. The "--force" option for `git submodule update` is only available in git 1.7.6 and above.

PR #722, opened 2015-06-25
Preserve rootfs permissions

badie:
Move the extraction of the cflinuxfs2 rootfs into the dea_next job vm. This
will ensure that the ownership of /home/vcap dir in the rootfs is vcap:vcap.
The compilation & distribution steps do not preserve the ownership and that's
why we are moving the extraction of the rootfs to the warden_ctl script.

Signed-off-by: Svett Ralchev <sralchev@pivotal.io>

### DIFF
--- a/jobs/dea_next/spec
+++ b/jobs/dea_next/spec
@@ -116,7 +116,7 @@ properties:
   dea_next.stacks:
     default:
       - name: "cflinuxfs2"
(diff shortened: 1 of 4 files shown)


dieucao:
Merged. 
https://github.com/cloudfoundry/cf-release/commit/96823da1e8e921de5b1b9d927f8289ed6c97d339

PR #727, opened 2015-07-03
Uaa local saml signing configurable

### LINKED ISSUE #707: Setting login.saml.signMetadata and login.saml.signRequest at login.yml
Hi.
I have been doing some testing integrations with a SAML identity provider. In order to achieve it, I had to set to false `login.saml.signMetadata` and `login.saml.signRequest` at `login.yml` UAA configuration file (I know that I would not do this in production, but it was just a quick test in a lab environment).
However, there are no properties defined in the UAA job of the cf-release to set these configuration settings in the UAA and I think they may be useful.

palonsoro:
Intended to fix #707


fhanik:
@palonsoro Commits look good, I would recommend squashing them into a single commit.

PR #734, opened 2015-07-14
extract resource_pools property to separate file

JamesClonk:
This allows to define custom resource_pools through additional spiff template files.
For example:

```yaml
meta:
  resource_pools:
    server_groups:
      etcd: (( merge ))
  stemcell: (( merge ))                                    

resource_pools:
  - <<: (( merge ))

  - name: etcd_z1
    cloud_properties:
      <<: (( merge ))
      instance_type: m1.medium
      scheduler_hints:
        group: (( meta.resource_pools.server_groups.etcd ))
    network: cf1
    stemcell: (( meta.stemcell ))
    env: (( merge || meta.default_env ))                   

jobs:
  - <<: (( merge ))

  - name: etcd_z1
    <<: (( merge ))
    resource_pool: etcd_z1
```
This way we can configure the openstack server-group (which is used for vm-anti-affinity) for each job separately.

Signed-off-by: Fabio Berchtold <fabio.berchtold@swisscom.com>

### DIFF
--- a/generate_deployment_manifest
+++ b/generate_deployment_manifest
@@ -16,6 +16,7 @@ fi
 
 spiff merge \
   $templates/cf-deployment.yml \
+  $templates/cf-resource-pools.yml \
   $templates/cf-jobs.yml \
   $templates/cf-properties.yml \
   $templates/cf-lamb.yml \
--- a/spec/fixtures/aws/cf-manifest.yml.erb
+++ b/spec/fixtures/aws/cf-manifest.yml.erb
@@ -657,12 +657,6 @@ meta:
   releases:
   - name: cf
     version: latest
-  stemcell:
-    name: bosh-aws-xen-hvm-ubuntu-trusty-go_agent
-    version: latest
-  default_env:
-    bosh:
-      password: $6$4gDD3aV0rdqlrKC$2axHCxGKIObs6tAmMTqYCspcdvQXh3JJcvWOY2WGb4SrdXtnCyNaWlrf3WEqvYR2MYizEGp3kMmbpwBC6jsHt0
 name: null
 networks:
 - name: cf1
--- a/spec/fixtures/openstack/cf-manifest.yml.erb
+++ b/spec/fixtures/openstack/cf-manifest.yml.erb
@@ -658,12 +658,6 @@ meta:
   releases:
   - name: cf
     version: latest
-  stemcell:
-    name: bosh-openstack-kvm-ubuntu-trusty-go_agent
-    version: latest
-  default_env:
-    bosh:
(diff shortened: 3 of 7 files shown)


JamesClonk:
Oops, just saw that there's some testing going on on the manifest/template files. Gotta fix those first.

Amit-PivotalLabs:
Hi @JamesClonk,

Thanks for your PR.  We are undertaking an effort as part of the new OSS Release Integration team to simplify manifest generation, unifying conventions across releases, and dovetailing the feature set with the BOSH CLI itself.  Better support for VM anti-affinity, for instance, is a planned feature in BOSH.

At the time, we are not planning to expand the usage of spiff and the current templates.  Current standard practice is to maintain separate post-processing tooling to add/modify things within the manifest after spiff merge.

I'm interested in understand your use case better though.  If you could describe at a higher level what you're trying to accomplish in your deployment, I'd like to be able to take them into consideration for our roadmap.

Thanks,
Amit

JamesClonk:
Hi @Amit-PivotalLabs,

OK, I see.  
What I wanted to achieve with this PR was to be able to define my own custom `resource_pools` through additional spiff templates.
Since currently all resource_pools are defined directly at the top in `cf-deployment.yml`, I cannot simply add another one "later" through a spiff merge, and I don't want to have to maintain any modifications on the files in cf-release/templates.
See my example template above for creating a custom etcd resource_pool. I would then define the template order like this in the `stub.yml`: 
```
templates:
- cf-release/cf-deployment.yml
- custom/my_etcd_pool.yml
- cf-release/cf-resource-pools.yml
```

drnic:
We've wanted to add more resource pools too - where the different jobs want to decouple from the small subset of instance types available

Amit-PivotalLabs:
@JamesClonk @drnic thank you for raising this need.  We are in the process of improving manifest generation within cf-release and across cloudfoundry in general, and one of the goals is absolutely the decoupling of the jobs from the limited pre-defined set of resource pools definitions in these templates.  We want to provide the flexibility to define cloud properties per job, similar to [how diego-release does it](https://github.com/cloudfoundry-incubator/diego-release/blob/develop/manifest-generation/misc-templates/iaas-settings.yml).  We want to greatly reduce the complexity in the current set of templates and stubs, and eventually move away from spiff altogether.

That said, I think this pull request is a reasonable refactoring of the existing templates.  Just a couple pieces of feedback:

* `cf-resource-pools` does not need to merge in `networks`, `jobs` or `properties`
* `cf-deployment` no longer needs to merge in `meta.stemcell` or `meta.default_env`.
* If you make the above change, the tests will fail -- the spec fixtures should be updated to no longer expect `meta.stemcell` and `meta.default_env`.

JamesClonk:
@Amit-PivotalLabs OK, changes made as suggested.

Amit-PivotalLabs:
@JamesClonk sorry, my mistake for merging this.  PRs should be made against the develop branch.  I will correct this, but just letting you know for future reference.

PR #751, opened 2015-07-29
[Power Support] Update build configuration files to support IBM Power platform.

allomov:
The problem is that some packages do need updated `config.guess` file to be build on IBM Power platform. 

The details and discussion about the way of implementation and reasons of this PR can be found in BOSH project: https://github.com/cloudfoundry/bosh/pull/898 and https://github.com/cloudfoundry/bosh/pull/892.


jpalermo:
Hi @allomov 

This pull request seems fine, but we have no way of verify this works and future support. Also, we cannot gate our releases for this specific architecture because we have no ci pipeline.

Could you move `src/config/config.guess` and `src/config/config.sub` to something like `src/architecture-support/config/...` for naming convenience.  

@DanLavine && @jpalermo,
OSS Release Integration Team

allomov:
@jpalermo I agree with all you say.

We decided to move changes for Power support to upstream step by step, to avoid large commits and manage this changes easier.

Currently we work on running CATs on CF deployed to Power. The main problem here is that we use special buildpacks to run apps on the CF, still default buildpacks are not designed to be run on Power. I guess after the work will be over, you'll be provided with the way to run tests.

allomov:
@jpalermo I made changes that you suggested. Could you please verify them?

PR #752, opened 2015-07-29
Checks if bundler version is correct in pre_packaging scripts.

allomov:
Hey, all.

I've noticed that `.gem` files that was packaged by bundler with the latest version not always can be found by the earlier version of bundler. To avoid such problems in I added checks in `pre_packaging` script of my release and it helped several people.

For the same purpose I created this PR https://github.com/cloudfoundry-incubator/bosh-openstack-cpi-release/pull/6.

### DIFF
--- a/packages/cloud_controller_ng/pre_packaging
+++ b/packages/cloud_controller_ng/pre_packaging
@@ -1,5 +1,10 @@
 set -e -x
 
+if [[ "`bundle --version`" != *"1.9"* ]]; then
+  echo "You should use bundler with version 1.9.x"
+  exit 2
+fi
+
 cd ${BUILD_DIR}/cloud_controller_ng
 BUNDLE_WITHOUT=development:test bundle package --all --no-install --path ./vendor/cache
 rm -rf ./vendor/cache/ruby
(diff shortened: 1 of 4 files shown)


jpalermo:
Hi @allomov,
This seems like a good idea, but the check is a bit too strict. We would want it to work with current versions of bundler (1.10.*) as well as future versions too.

allomov:
@jpalermo I think that 1.10 can be added to this checks, but I am sure that future versions of bundler can break something in the same manner. This checks will help to avoid errors that are hard to debug.

As I see there is another way to cope with such errors: `pre_packaging` scripts should be run in determined environment, using only its dependencies. But this breaks the idea behind `pre_packaging` scripts.

jpalermo:
@allomov Adding code to alert users that their bundler is too old to work is ideal.

But at the same time, we don't believe that newer versions of bundler are likely to break the functionality, and we don't want to create a situation where we have to update the pre-packaging scripts whenever a new version of bundler comes out.

jpalermo:
Closing due to inactivity. Feel free to reopen though.

PR #758, opened 2015-08-04
Fix 3.19 kernel issue with SYN's being silently dropped when tcp_tw_r…

MarkKropf:
http://serverfault.com/questions/342741/what-are-the-ramifications-of-setting-tcp-tw-recycle-reuse-to-1

When you enable tcp_tw_recycle, the kernel becomes much more aggressive, and will make assumptions on the timestamps used by remote hosts. It will track the last timestamp used by each remote host having a connection in TIME_WAIT state), and allow to re-use a socket if the timestamp has correctly increased. However, if the timestamp used by the host changes (i.e. warps back in time), the SYN packet will be silently dropped, and the connection won't establish (you will see an error similar to "connect timeout"). If you want to dive into kernel code, the definition of tcp_timewait_state_process might be a good starting point.

### DIFF
--- a/jobs/dea_next/templates/dea_ctl.erb
+++ b/jobs/dea_next/templates/dea_ctl.erb
@@ -46,7 +46,7 @@ case $1 in
     # incorrectly states the default as enabled. It can be changed to 1 (enabled) in many cases. Known to cause some
     # issues with hoststated (load balancing and fail over) if enabled, should be used with caution.
 
-    echo 1 > /proc/sys/net/ipv4/tcp_tw_recycle
+    echo 0 > /proc/sys/net/ipv4/tcp_tw_recycle
 
     # TCP_TW_REUSE
     # This allows reusing sockets in TIME_WAIT state for new connections when it is safe from protocol viewpoint.
--- a/jobs/gorouter/templates/gorouter_ctl.erb
+++ b/jobs/gorouter/templates/gorouter_ctl.erb
@@ -44,7 +44,7 @@ case $1 in
         # incorrectly states the default as enabled. It can be changed to 1 (enabled) in many cases. Known to cause some
         # issues with hoststated (load balancing and fail over) if enabled, should be used with caution.
 
-        echo 1 > /proc/sys/net/ipv4/tcp_tw_recycle
+        echo 0 > /proc/sys/net/ipv4/tcp_tw_recycle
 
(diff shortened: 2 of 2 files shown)


Amit-PivotalLabs:
@fraenkel @sykesm: I'm told by @MarkKropf that you have context on why `tcp_tw_recycle` was originally being set to 1.  Can you give feedback on this PR?

@MarkKropf, perhaps you could add some details about what issue this fixes, and the types of environments in which this problem is reproducible.

MarkKropf:
In stemcells with the 3.19 kernel, we found that the default windows 7 TcpWindowSize causes timestamps on packets received by gorouters behind hardware loadbalancers to come in out of sync and be dropped. The windows 7 TcpWindowSize can be changed from the defaults as a work around, however it is commonly recommended to not enable both tcp_tw_recycle and tcp_tw_reuse at the same time. 

comments with the original commit
image

sykesm:
@Amit-PivotalLabs: We were probably a bit over zealous. When we were looking through the descriptions of what recycle did and tried to correlate it to known use cases where we had ha proxy or an ELB in front of the router, we didn't think it would be a problem.

Since someone has a hit a problem with windows, setting the flag back to 0 seems like the right thing to do.

Amit-PivotalLabs:
@sykesm @MarkKropf thanks!

youngm:
Wow!

This may be the solution to some dropped connection and 3 second connection delay problems we've been seeing during load tests.  I'm going to try this in our deploy next week.

youngm:
Any thoughts on removing this from the DEA network turning config also?  I've been battling with DEA dropping SYN packets as well as the router.  Though I don't have as good an explanation as @MarkKropf regarding windows and hardware load balancers.  If you want me to provide a better case before changing this setting I'll understand.

https://github.com/cloudfoundry/cf-release/blob/master/jobs/dea_next/templates/dea_ctl.erb#L49

Amit-PivotalLabs:
Hi @youngm 

That was exactly the line that was changed by this PR.  See [this commit](https://github.com/cloudfoundry/cf-release/commit/d10b17287709aa5717345fec45015d0e5acf0cfb).  It just hasn't made its way to master branch yet.

youngm:
@Amit-PivotalLabs Oh great!  I missed the additional commit for DEA also.  Thanks.

PR #772, opened 2015-08-25
Improve manifest generation scripts

ematpl:
bosh-lite/make_manifest:
* remove perl dependency
* avoid making temp dir and temp stub
* avoid changing working directory
* tolerate whitespace in paths
* fix whitespace and unusual indentation
* improve variable cohesion

bosh-lite/cf-stub-spiff.yml
* add explicit spiff 'merge' directive for director_uuid

scripts/generate_deployment_manifest:
* tolerate whitespace in paths

generate_deployment_manifest:
* deprecation notice refers to deprecated file

### DIFF
--- a/bosh-lite/cf-stub-spiff.yml
+++ b/bosh-lite/cf-stub-spiff.yml
@@ -1,6 +1,5 @@
 ---
 name: cf-warden
-director_uuid: PLACEHOLDER-DIRECTOR-UUID
 releases:
   - name: cf
     version: latest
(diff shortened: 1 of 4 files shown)


amitkgupta:
Thanks, looks great!  So happy to see the perl gone.

- Does the change to `scripts/generate_deployment_manifest` need bash to tolerate the whitespace?  
- Does the `cf-stub-spiff.yml` even need the director UUID merge?  I believe the top level template already has that, and I'd prefer to avoid things called "stub" having spiff (( merge )) in them.

ematpl:
- Nope, we're not sourcing `scripts/generate_deployment_manifest`, so `$0` should be fine.
- Nope, the `director_uuid` line can be removed entirely.

PR #773, opened 2015-08-26
Add router.extra_headers_to_log to spec, template and spiff template.

simonjohansson:
More info here https://github.com/cloudfoundry/gorouter/pull/92

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -76,6 +76,9 @@ properties:
   router.logrotate.size:
     description: "The size at which logrotate will decide to rotate the log file"
     default: 2M
+  router.extra_headers_to_log:
+    description: "A list of headers that log events will be annotated with"
+    default: []
 
   nats.user:
     description:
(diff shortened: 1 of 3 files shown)


leochu:
Hi Simon,
Would you please change the jobs template from `extra_headers_to_log: (( merge || [] ))` to `extra_headers_to_log: (( merge || nil ))` ?

The convention is to have the default value set in the job spec.  In this case, the empty slice is overriding the value in the job spec which would cause confusion when a user tries to change the default value in the job spec.

It would be nice if you could squash your commits for this PR.

I've also commented on https://github.com/cloudfoundry/gorouter/pull/92/files where a space is logged in the case where extra_headers_to_log is set to an empty slice, which is the default value.

I'm happy to merge both PRs once these two comments are addressed.

Thanks,
Leo

shalako:
This was merged with commit https://github.com/cloudfoundry/cf-release/commit/c20b1693c58676509db89944b194b2fd9698497f.

Shipped with cf-release v217

PR #775, opened 2015-08-28
Change UAA timestamp logging to be rfc3339 compatible

youngm:
This will add timezone information and improve timestamp detection on log servers like Splunk.

### DIFF
--- a/jobs/uaa/spec
+++ b/jobs/uaa/spec
@@ -258,6 +258,9 @@ properties:
   uaa.logging_level:
     description: Set UAA logging level.  (e.g. TRACE, DEBUG, INFO)
     default: DEBUG
+  uaa.logging_use_rfc3339:
+    description: Sets the time format for log messages to be rfc3339 compatible. 
+    default: false
   login.protocol:
     description: "Scheme to use for HTTP communication (http/https)"
   login.ldap.profile_type:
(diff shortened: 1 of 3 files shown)


Amit-PivotalLabs:
@mbhave @fhanik @sreetummidi just want to check in, do you see any issue with this PR?  Does anything else depend on the existing log format?

sreetummidi:
Looks good to us

Thanks,
Sree Tummidi
Sr. Product Manager
Identity - Pivotal Cloud Foundry

sreetummidi:
@amitkgupta It actually does have implications on the customers using the existing log format.
If there are SIEM/Log-Collection/Analytics solutions hooked up which expect this format for parsing, they will break.
We should make this a configurable property in the manifest instead so that it can be set to any required format
-Sree

Amit-PivotalLabs:
Thanks @sreetummidi 

@youngm can you make the log format a manifest-configurable property?

youngm:
@Amit-PivotalLabs Sure, I'll update the PR.

youngm:
@Amit-PivotalLabs I've updated the PR to make it optional and I updated varz logging as well.

fhanik:
Wouldn't the smoothest to just expose the pattern? Or is that risky to break the logging?

uaa:
  logging: 
    format: yyyy-MM-dd'T'HH:mm:ss.SSSXXX

youngm:
@fhanik I'd be willing to do that also if you want to revert this. I mostly just need timezone specified and rfc3339 seemed to be the standard that includes timezone.

However, I think it would be nice to establish a consistent format for date in logs rather than any format the user wants.  As I find other components not using rfc3339 or a epoch timestamp I plan to submit PRs to those projects as well to help provide some uniformity in this space.

PR #777, opened 2015-09-01
Fix description of staging_upload_user and staging_upload_password

mmc2004jp:
These properties should be credential for Cloud Controller's upload endpoint (not S3 or other IaaSs credential).

I took the description from the following article:
http://docs.cloudfoundry.org/deploying/cf-stub-openstack.html

> Replace the staging_upload_user: username with the account user name used to upload files to the Cloud Controller. 
> Replace the staging_upload_password: password with the password of the account used to upload files to the Cloud Controller.

### DIFF
--- a/jobs/cloud_controller_clock/spec
+++ b/jobs/cloud_controller_clock/spec
@@ -186,10 +186,10 @@ properties:
 
   cc.staging_upload_user:
     default: ""
-    description: "S3 Access key for staging droplets on AWS installs; Blobstore user for other IaaSs"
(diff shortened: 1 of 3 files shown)


njbennett:
Hi @mmc2004jp,

Thanks for the PR. The description that's there does need to be changed, but it would be nice if the new description made it clear that this is for internal use only.

Thanks,
@njbennett & @DanLavine, CF OSS Release Integration Team

mmc2004jp:
Hi @njbennett & @DanLavine,

That's good point. I changed the description.
Let me know if you have any suggestions/improvements.

Thanks,

PR #781, opened 2015-09-04
Add support for specifying an existing user in CATs.

youngm:
In our deployments we use LDAP.  Exposing the existing user attribute allows us to run the CATs with an existing LDAP test account.

### DIFF
--- a/jobs/acceptance-tests/spec
+++ b/jobs/acceptance-tests/spec
@@ -78,3 +78,7 @@ properties:
     description: Timeout for long curls
   acceptance_tests.broker_start_timeout:
     description: Timeout for broker starts
+  acceptance_tests.existing_user:
+    description: "The username of an existing user. If set, the acceptance-tests will push apps and perform other actions as this user, otherwise its default behaviour is to create a temporary user for such actions."
+  acceptance_tests.existing_user_password:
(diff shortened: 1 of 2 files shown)


jobs/acceptance-tests/spec
```diff
@@ -78,3 +78,7 @@ properties:
     description: Timeout for long curls
   acceptance_tests.broker_start_timeout:
     description: Timeout for broker starts
+  acceptance_tests.existing_user:
+    description: The username of an existing regular user.
```
Amit-PivotalLabs: Suggested change:

"The username of an existing user.  If set, the acceptance-tests will push apps and perform other actions as this user, otherwise its default behaviour is to create a temporary user for such actions."

jobs/acceptance-tests/spec
```diff
@@ -78,3 +78,7 @@ properties:
...
+  acceptance_tests.existing_user_password:
+    description: The password of an existing regular user.
```
Amit-PivotalLabs: Suggested change:

"The password of the existing user.  Only required if the existing user property is also being set."

youngm:
@Amit-PivotalLabs Good suggestions.  I've updated the commit with your suggestions exactly.

PR #784, opened 2015-09-11
Add enable_routing_api property to cc spec file and cf-jobs template.

atulkc:
This PR is to add router.enable_routing_api property in CC spec file to enable conditionally exposing routing api endpoint url in /v2/info.

Tracker: [#100974596]

Regards,
@atulkc and @fordaz

### DIFF
--- a/templates/cf-jobs.yml
+++ b/templates/cf-jobs.yml
@@ -694,6 +694,7 @@ properties:
 
   router:
     enable_ssl: (( merge || nil ))
+    enable_routing_api: (( merge || true ))
     ssl_cert: (( merge || nil ))
     ssl_key: (( merge || nil ))
     cipher_suites: (( merge || nil ))
--- a/jobs/cloud_controller_ng/spec
+++ b/jobs/cloud_controller_ng/spec
@@ -370,6 +370,10 @@ properties:
     description: "Port for doppler_logging_endpoint listed at /v2/info"
     default: 443
 
+  router.enable_routing_api:
+    description: "Whether to expose the routing_api endpoint listed at /v2/info"
+    default: true
+
   cc.db_encryption_key:
     default: ""
     description: "key for encrypting sensitive values in the CC database"
--- a/spec/fixtures/aws/cf-manifest.yml.erb
+++ b/spec/fixtures/aws/cf-manifest.yml.erb
@@ -991,6 +991,7 @@ properties:
       rotate: 5
       size: 3M
     enable_ssl: true
+    enable_routing_api: true
     extra_headers_to_log: null
     ssl_cert: |
       -----BEGIN CERTIFICATE-----
--- a/spec/fixtures/openstack/cf-manifest.yml.erb
+++ b/spec/fixtures/openstack/cf-manifest.yml.erb
@@ -983,6 +983,7 @@ properties:
       rotate: 5
       size: 3M
     enable_ssl: true
+    enable_routing_api: true
     extra_headers_to_log: null
     ssl_cert: |
       -----BEGIN CERTIFICATE-----
--- a/spec/fixtures/vsphere/cf-manifest.yml.erb
+++ b/spec/fixtures/vsphere/cf-manifest.yml.erb
@@ -998,6 +998,7 @@ properties:
       rotate: 5
       size: 3M
     enable_ssl: true
+    enable_routing_api: true
     extra_headers_to_log: null
     ssl_cert: |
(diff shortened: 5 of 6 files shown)


atulkc:
This PR is part of changes required to add routing api endpoint url in /v2/info. The other set of changes to CC are submitted via https://github.com/cloudfoundry/cloud_controller_ng/pull/432 PR.

templates/cf-jobs.yml
```diff
@@ -694,6 +694,7 @@ properties:
 
   router:
     enable_ssl: (( merge || nil ))
+    enable_routing_api: (( merge || true ))
```
Amit-PivotalLabs: Given `router.enable_routing_api` defaults to `true` in the job spec, can we avoid having this default value also in the manifest template?

Amit-PivotalLabs:
@atulkc @fordaz ,

What does it mean to expose the routing API endpoint in `/v2/info` if the routing API and/or consul are not actually deployed?

templates/cf-jobs.yml
fordaz: @Amit-PivotalLabs we added this in manifest template to control `enable_routing_api` property on both gorouter and cloud_controller_ng jobs. Otherwise we end up changing the property defaults in two spec files for each jobs, which can get out of sync.

fordaz:
@Amit-PivotalLabs routing API endpoint will get exposed in `/v2/info` if enable_routing_api is `true`. This means if routing api is not deployed then this property must be set to `false` and it won't appear in `/v2/info`. We are not sure what you mean by consul deployment and its relation to routing api.

Regards,
@fordaz and @atulkc

Amit-PivotalLabs:
@fordaz @atulkc 

My understanding was that routing-api has some dependency on consul.  We're saying routing-api is now enabled by default, but consul servers are not part of the deployment by default.  How does that work?

templates/cf-jobs.yml
Amit-PivotalLabs: I see.  That's unfortunate, hopefully this will be solved once [BOSH links](https://github.com/cloudfoundry/bosh-notes/blob/master/links.md) land.  In the mean time, I'd prefer to avoid increasing the number of places where defaults are set, and would prefer the risk of the CC and router properties' defaults going out of sync.  This doesn't fully solve that problem anyways, in the case where the stub explicitly specifies `enable_routing_api: null`.

atulkc:
Routing api does register with consul and gorouter uses consul for discovering routing api. Gorouter uses enable_routing_api property to decide whether to use routing api. So in case where enable_routing_api property is set to true (which is default) and routing api is not deployed gorouter will try to subscribe to events from routing api and will fail. It then just logs an error and moves on.

Amit-PivotalLabs:
Okay, here's my guess as to how all this works and what it is for:

If routing_api_enabled is true, gorouter attempts to discover routing_api via consul DNS.  If Routing API is actually running, it will attempt to register itself with consul DNS, and it will also add `api.SYSTEM_DOMAIN/routing` to its database.  If consul is running, gorouter will discover Routing API and sync any routes that had been registered with the Routing API (including the Routing APIs own public api.SYSTEM_DOMAIN/routing route) with gorouter's route table.  Other internal components will eventually be updated to register routes with the Routing API after discovering it over consul DNS; right now they register with the gorouter over NATS.  External users, and applications running on the system, can also use the Routing API.  They discover the Routing API from /v2/info.  Assuming they are authenticated, they can register routes with the API, and the gorouter will sync those too.  `api.SYSTEM_DOMAIN/routing` works, and doesn't clash with CC, because of this logic: https://github.com/cloudfoundry/gorouter/blob/master/registry/trie.go#L45.

If all that is true, then the current default is:

* no consul running
* so gorouter doesn't discover Routing API
* so gorouter doesn't add api.SYSTEM_DOMAIN/routing to its route table
* but by default, routing_api_enabled is true, so we advertise this URL in /v2/info
* so by default, we advertise a URL that does not work

Correct?

atulkc:
@Amit-PivotalLabs 

You are correct! While `enable_routing_api` default of true works for gorouter, it doesn't for CC. I have asked @shalako and @crhino to connect with you to address how we want to resolve this.

shalako:
@Amit-PivotalLabs we'll make the property unconditional and resubmit.

Amit-PivotalLabs:
Since this PR has diverged a fair bit now, do you guys mind submitting a new PR?  It will also be quicker for me to prioritize because I'll see the new PR in my icebox.

PR #785, opened 2015-09-12
Modify DEA configuration for higher performance

rexxar-liang:
- nf_conntrack_max is set to 65536 by default, we found linux kernel
  drop tcp handshake request with the message "nf_conntrack: table
  full, dropping packet" by using command "dmesg"
- in order to support higher performance, we set nf_conntrack_max
  to a larger number

Signed-off-by: Yun Lv <lvyun@huawei.com>

### DIFF
--- a/jobs/dea_next/templates/dea_ctl.erb
+++ b/jobs/dea_next/templates/dea_ctl.erb
@@ -53,6 +53,12 @@ case $1 in
     # Default value is 0 (disabled). It is generally a safer alternative to tcp_tw_recycle
 
     echo 1 > /proc/sys/net/ipv4/tcp_tw_reuse
+
+    # NF_CONNTRACK_MAX
+    # Default value is 65536. We set it to a larger number in the purpose for higher performance.
+    # We do not need to worry about DDOS attack because there are other components(gorouter,etc) in front of DEA will handle it.
+
+    echo 1024000 > /proc/sys/net/nf_conntrack_max
     <% end %>
 
     exec /var/vcap/packages/ruby-2.1.7/bin/ruby /var/vcap/packages/dea_next/bin/dea \


Amit-PivotalLabs:
Thanks @rexxar-liang, we plan on addressing this in the stemcell itself:

https://www.pivotaltracker.com/story/show/103144524

I will close out this PR.

Thanks,
Amit

rexxar-liang:
If we addressing this in the stemcell, I think there are some cons, what is your suggestions?
1、Not all stemcell will have this file("/proc/sys/net/nf_conntrack_max")
     It seems that only if the IP tables is install and configured, then the file will be there.
     So whether it means we will have two different stemcell?
2、Because stemcell is a general solution for all VM, but some VMs used by many other components(like haproxy, gorouter) which do not need it.

Amit-PivotalLabs:
@rexxar-liang 

That's a good point, let me investigate.  We will definitely not have two stemcells.  If we put this in the job as your PR does, we were actually thinking to decrease the time window rather than increase the max connections: `sysctl net.netfilter.nf_conntrack_tcp_timeout_time_wait=10` (the original default is 120).  Thoughts about that?

We could also separately increase the max conntrack, but thought shrinking the time window would be the safer of the two settings as far as which one we try first.

rexxar-liang:
@Amit-PivotalLabs 
Thanks for your response. 
We will have a test as your suggestion in our performance test and Synchronize the result to you.

Amit-PivotalLabs:
Great, looking forward to hearing about it.

rexxar-liang:
@Amit-PivotalLabs 
We tried only use 'sysctl net.netfilter.nf_conntrack_tcp_timeout_time_wait=10', we still can see linux kernel drop tcp handshake request.

Actually, after gorouter get the respond from DEA, gorouter will send a FIN to DEA and release the connection immediately, so the connection will not keep for a long time, that is why "nf_conntrack_tcp_timeout_time_wait" will not help us.

So far, we found the only solution is to set nf_conntrack_max to a larger number.
Do you have any other suggestion?

Amit-PivotalLabs:
@rexxar-liang thanks.

So turns out decreasing `nf_conntrack_tcp_timeout_time_wait` didn't work for you, huh?

Are there any implications on memory utilization by increasing the `nf_conntrack_max` number?  I've also seen it suggested to increase `net.netfilter.nf_conntrack_buckets`, I'm not sure what the tradeoffs are.

I'd love to hear thoughts from @dsboulder @b @thansmann.

Thanks,
Amit

sgpivotal:
Hi Amit,

Would increasing the number of buckets be an option rather than increasing the contrack_max?
I think the number of buckets may be set quite low in relation to the amount of memory available.

i.e.
net.netfilter.nf_conntrack_buckets = 65536

https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt

Regards,
Shane

rexxar-liang:
Hi Shane & Amit,
    nf_conntrack_buckets is read only, and we can not modify.
    I found that for both physical machine and VMs bring by unbuntu stemcell, nf_conntrack_buckets is 16384, and nf_conntrack_max is 65536, which is not enough for high performance scenario.
    ```For systems with more than 4GB of memory it will be 65536 buckets``` mentioned in [link](https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt) is not the case in fact.

Amit-PivotalLabs:
@rexxar-liang I tried re-opening this PR, but it was against `master`.  We only do PRs against `develop`, so I will just ask the core development team to make this change.

rexxar-liang:
@Amit-PivotalLabs Thanks for your reply. We can do that, I will raise the PR against `develop` as you wish.
About the value of nf_conntrack_max, I think 65536 * 4 will be fine. because the memory of DEA vm is usually greater than 4G, so we use the default value suggested in [link](https://www.kernel.org/doc/Documentation/networking/nf_conntrack-sysctl.txt), and nf_conntrack_max should be 65536 * 4, that is 262144.
Do you have any suggestions?

Amit-PivotalLabs:
Hey @rexxar-liang 

No need to submit a PR, I'll have the core dev team make the change directly.  You can see the story prioritized here: https://www.pivotaltracker.com/story/show/104485742

It's not clear what a reasonable number for `nf_conntrack_max` is.  The documentation is quite confusing.  For instance, the default for `nf_conntrack_buckets` it says "divide total memory by 16384", but total memory in what units?  It also seems contradictory, saying that the value is limited to 16384, but then in the next sentence says for systems with 4GB of memory, it's 65536.

It's also weird you had issues when trying to decrease `nf_conntrack_tcp_timeout_time_wait` to 10.  We tried this ([see here](https://www.pivotaltracker.com/story/show/102698920)) and saw no issues.  Also, the DEA and Gorouter set `tcp_fin_timeout` to 5s:

* https://github.com/cloudfoundry/cf-release/blob/68a2ec6769cfcc0e0a2c58fc1c3ce79ce595018a/jobs/dea_next/templates/dea_ctl.erb#L42
* https://github.com/cloudfoundry/cf-release/blob/5fa14702bca4d36d1fdc7241c63d0b3e40dcbe90/jobs/gorouter/templates/gorouter_ctl.erb#L40

So I would have guessed it safe to set `nf_conntrack_tcp_timeout_time_wait` to 10, but I'm not sure.

PR #786, opened 2015-09-17
Add better error message on missing infrastructure arg

tushar-dadlani:


Amit-PivotalLabs:
Hey @tushar-dadlani,

Thanks for the PR.  What behaviour does this PR change?  I think you want it to check, if the first argument is not provided, or if it's provided but isn't one of the supported infrastructures, then print the usage and exit.  Isn't this already the current behaviour?

tushar-dadlani:
@Amit-PivotalLabs Looks like it was a mistake on my end. Earlier today, I tried to run a generate-deployment-manifest with empty args and saw that it didn't work as expected. After checking the current behavior on top of develop, it is correct. I possibly had a stale local copy.

PR #788, opened 2015-09-18
Commit script no alias

markstgodard:
Hi MEGA team

This week I updated our Routing team CI pipeline to auto-bump cf-release
however the docker image we use does not have the standard
aliases (i.e. from sprout wrapped machines)

I am able to work around by adding the git alias manually in one of my pipeline scripts,
https://github.com/cloudfoundry/deployments-routing/blob/master/routing-ci/scripts/ci/bump
but ideally, don't like having to :)

Cheers


markstgodard:
Closing, incorrectly created off master vs. develop.

PR #789, opened 2015-09-18
Change commit_with_shortlog to not use git alias

markstgodard:
Allow ./scripts/commit_with_shortlog to be Concourse pipeline friendly

We updated our Routing CI pipeline to auto-bump, however the docker image we use does not have the standard aliases (i.e. from sprout wrapped machines).
Its not a major issue and can be worked around it by creating the git alias in concourse build script  but I thought others teams may encounter if they use ./commit_with_shortlog

Mark - CF Routing Team

### DIFF
--- a/scripts/commit_with_shortlog
+++ b/scripts/commit_with_shortlog
@@ -1,4 +1,4 @@
 #!/bin/bash
 script_dir=$(dirname $0)
 
-$script_dir/staged_shortlog | git ci -F -
+$script_dir/staged_shortlog | git commit -F -


sykesm:
`ci` is a common alias for `duet-commit` on the workstations. That's the tool we use for pairing to include multiple contributors in the commit. The expectation is that `commit_with_shortlog` is done by programmers as part of ceremony involved with bumping submodules in a release.

I think it would be surprising if this script stopped using `ci` for commits.

markstgodard:
Hey Matt,

Yeah `ci` is a definite common git alias (I have it) and the standard CF pairing workstations have it,  however the only minor problem I encountered this week was that I setup the CF Routing Team's Concourse pipeline to "auto-bump" if all the unit tests, acceptance tests pass...  and I still wanted to use the same commit_with_shortlog script when bumping.
i.e. our routing-ci bot will do the auto-bump...
https://github.com/cloudfoundry/cf-release/commit/83b8ae5e20d2929dd1743092c62d5cabdc5300cb 

However when calling the script in a Concourse job there is no git alias for `ci`.
Since the Concourse jobs use docker containers, they are not aware of the git aliases we typically use...
so yeah I just added the git alias in the build script... but it felt wrong :)

I also expect if other CF teams setup their Concourse pipelines to auto-bump cf-release, then they would encounter the same issue.

So yeah I totally agree..  `ci` is very common.. I never use `commit`,  however I wanted to use `commit_with_shortlog` in our Concourse build, for consistency in commit messages, etc.  

Maybe I am the first person to use this script in a Concourse pipeline ?  :)

Cheers

sykesm:
The way I look at it, either the pipeline does something to enable use of the script or all of the developers that are currently using the script will need to write new scripts to do the right thing. Given `commit_with_shortlog` is a one line wrapper around `staged_shortlog`, I think the path of leas resistance would be for you to use that instead to generate your commit message.

zaksoup:
I agree with @sykesm, this change breaks expectations that all of our developers have about the behavior of commit_with_shortlog. Is there a reason that the line in your build script can't literally be
`$script_dir/staged_shortlog | git commit -F -`?

Best,
@zaksoup & @medvedzver

markstgodard:
@zaksoup Yeah I can work around,   its not a problem.

The only thing I wanted to prevent was other CF teams which plan to auto-bump using `commit_with_shortlog` in a Concourse pipeline will encounter this and the discussion may come up again

Since (as CF Developers) we have specific prerequisite environment setup to do CF dev  (i.e. specific git aliases),  then I guess technically it should be the onus of the Concourse job to adhere to those prereqs.

So if I treat my Concourse job worker like a CF Developer and setup the prereqs.  In my case, I worked around by adding the git alias so I could use `commit_with_shortlog`.   This way (if for some reason) `commit_with_shortlog` changes in cf-release, then I am not out of sync.   I'll just have to make sure my Concourse job has the CF dev prerequisite setup.

Anyway, PR close away :)

Cheers

PR #790, opened 2015-09-20
Do not require use of static_ips or subnets in spiff templates

keymon:
Currently the `templates/cf-jobs.yml` template requires that the network subnets and the `static_jps` for several jobs (nats, etcd, nfs, consul, etc.) are defined.

But using alternative CPIs, like the new [Google Compute BOSH CPI](https://github.com/frodenas/bosh-google-cpi/) static IPs and subnets these values are not currently supported. 

With this PR it is possible to create a new `templates/infrastructure_gce.yml` where no static IPs are defined  (we will summit an example soon). In the stub files we can then overwrite the required values (nats, etcd, consul and NFS) to use BOSH DNS names (or other alternative service discovery).

This PR does not change the previous behaviour if the static IPs are defined and the new optional values not.  For example, for warden:

```
git checkout master
./scripts/generate_deployment_manifest warden \
    <(echo -e "---\ndirector_uuid: UUID") \
    ./templates/cf-minimal-dev.yml > \
        /tmp/warden_orig.yml

git checkout cf_jobs_without_static_ips_dependencies_latest
./scripts/generate_deployment_manifest warden \
    <(echo -e "---\ndirector_uuid: UUID") \
    ./templates/cf-minimal-dev.yml > \
        /tmp/warden_new.yml

diff /tmp/warden_orig.yml /tmp/warden_new.yml
```

### DIFF
--- a/templates/cf-jobs.yml
+++ b/templates/cf-jobs.yml
@@ -12,11 +12,13 @@ meta:
 
   consul_servers: (( merge || jobs.consul_z1.networks.cf1.static_ips jobs.consul_z2.networks.cf2.static_ips ))
 
+  nfs_client_ranges:
+    - (( .networks.cf1.subnets.[0].range || nil ))
+    - (( .networks.cf2.subnets.[0].range || nil ))
+
   nfs_server:
-    address: (( jobs.nfs_z1.networks.cf1.static_ips.[0] || nil ))
-    allow_from_entries:
-      - (( .networks.cf1.subnets.[0].range || nil ))
-      - (( .networks.cf2.subnets.[0].range || nil ))
+    address: (( merge || jobs.nfs_z1.networks.cf1.static_ips.[0] || nil ))
+    allow_from_entries: (( merge || meta.nfs_client_ranges ))
     share: ~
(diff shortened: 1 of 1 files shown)


keymon:
Here you can find an example of a [`infrastructure_gce.yml`](https://github.com/alphagov/cf-terraform/blob/normalise_manifests/manifests/templates/gce/cf-infrastructure-gce.yml) that works with this PR. 

We are working in [this codebase](https://github.com/alphagov/cf-terraform/tree/normalise_manifests/manifests) to get manifest for GCE similar to the AWS deployment using the official templates. Run it with `CF_RELEASE_PATH=../../cf-release generate_deployment_manifest.sh gce`.

**Note that this branch is WIP and the files might change**

zaksoup:
Hey @keymon,

These changes and requirements seem reasonable and we're going to review them and get back to you soon. I just want to warn you that the GCE CPI is not officially supported by the BOSH team and our ability to help with issues you may come across with it is going to be limited.

Best,
@zaksoup & @medvedzver
CF Release Integration

PR #791, opened 2015-09-21
Allow override default network for router jobs

keymon:
This PR allows override the network default name and configuration (`cf1` or `cf2`) for the `router_z1` and `router_z2` jobs and their corresponding resource pools.

This feature is required in the case that a different network is needed to configure the routers' loadbalancer. For instance,  in the case of the [BOSH Google CPI](https://github.com/frodenas/bosh-google-cpi), the automatic loadbalancer registering requires the definition of alternate networks with a custom `cloud_properties.target_pool` entry.

The implementation is similar to the `ha_proxy_z1`, where a new yaml node `default_network` is defined in each job, and included if the networks are not defined in the stubs. 

This PR is related to #790, as it enables deploy Cloud Foundry using the official templates from this repository on Google Compute Engine.

### DIFF
--- a/templates/cf-jobs.yml
+++ b/templates/cf-jobs.yml
@@ -494,9 +494,10 @@ jobs:
     templates: (( merge || meta.router_templates ))
     instances: 1
     resource_pool: router_z1
-    networks:
+    default_networks:
       - name: cf1
-        static_ips: (( merge ))
+        static_ips: ~
+    networks: (( merge || default_networks ))
     properties:
       consul:
         agent:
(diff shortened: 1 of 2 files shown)


zaksoup:
Hey @keymon,

Again, this seems reasonable. As I said for PR #790 we don't officially support the GCE CPI etc etc.

@zaksoup & @medvedzver

zaksoup:
Hey @keymon

We've pulled in your manifest changes. It's great that you're contributing! In the future please be aware that we have a series of unit tests that run against our manifest templates to ensure we do not break manifest generation. You can run these tests in the cf-release dir with `bundle exec rspec`. In this case the addition of the key `default_networks` failed these tests as they found an unexpected key. The fixtures that these tests run against can be found in `cf-release/spec/fixtures/[INFRASTRUCTURE]/cf-manifest.yml.erb`.

Thanks again for submitting this PR!

@zaksoup & @medvedzver

PR #792, opened 2015-09-21
Configure https as default in templates

sykesm:
Prefer ssh where possible for defaults. Without this, we have defaults which advertise a mix of secure and insecure endpoints to cli which can result in credential leaks when interacting with the UAA.

### DIFF
--- a/jobs/uaa/templates/login.yml.erb
+++ b/jobs/uaa/templates/login.yml.erb
@@ -21,9 +21,8 @@ logging:
   end
 %>
 
-<% protocol = (properties.login && properties.login.protocol) ? properties.login.protocol : "http" %>
+<% protocol = (properties.login && properties.login.protocol) ? properties.login.protocol : "https" %>
 <% if !properties.login || !properties.login.uaa_base
-  # Fix this to https when SSL certs are working in dev and staging
   uaa_base = "#{protocol}://uaa.#{properties.domain}"
 else
   uaa_base = properties.login.uaa_base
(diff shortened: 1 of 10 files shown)


Amit-PivotalLabs:
Thanks for the PR @sykesm 

Would the defaults you're changing make more sense to be made in the job specs rather than the job and manifest templates (and remove the hardcoded stuff in the various templates?)

What are the implications for the OSS pipeline configurations?  Will these changes require installing certs where our DNS for system and app domains terminate?  Will they have to be real certs, for UAA/Login to be happy?  We could also overwrite the defaults in our own manifest generation, but would prefer to err on the side of better security than not in our integration environments.

Also want to bring to your attention that we'd soon like to switch from the current uaa job to submoduling and symlinking the [uaa-release](https://github.com/cloudfoundry-incubator/uaa-release).  It has better defaults around security.

sykesm:
@Amit-PivotalLabs I'm happy to move the protocol default into the spec if you want me to rework this.

As for the pipelines, the changes to warden worked fine locally. Given the https protocol references in `cf-properties.yml`, I jumped to the conclusion that we were expecting the router/haproxy tier to support ssl at this point.

I really don't know what the impact will be to vsphere.

Amit-PivotalLabs:
Closing this PR in favour of [this commit](https://github.com/cloudfoundry/cf-release/commit/55cd6ac43fba45dfe9885efaa605a3f3230a2a90) which puts the defaults right down in the job specs.

sykesm:
Thanks!

PR #796, opened 2015-09-24
Fix logrotate setting of gorouter

kei-yamazaki:


leochu:
Hi Kei,
Nice catch!  This was already fixed in the commit below earlier this week.
https://github.com/cloudfoundry/cf-release/commit/4a2a307a333362d3b0d3dbbd8e792139196d958a

Thanks,
Leo
CF Routing Team

PR #798, opened 2015-09-29
Bump the open files for DEA and Warden

fraenkel:
Connections are closed when no longer used which means that there will be
a minimum of 5 * 256 connections on either side in the worst case.


fraenkel:
This is only to be used if and when we finally merge the DEA/Warden close connection branches.

PR #799, opened 2015-09-29
Update default AWS instance types

mtekel:
Use more appropriate AWS instance types. Jobs of the CF are well suited for AWS T2 bursty instance type. Runners use r3 instance with sustained performance and more memory capacity for apps. This update also decreases running cost for default AWS deployment.

```
small:  c3.large -> t2.small
medium: c3.large -> t2.medium
large:  m3.large -> t2.large
runner: m3.large -> r3.large
router: c3.large -> t2.medium
compilation:  c3.large -> t2.medium
small errand: c3.large -> t2.medium
large errand: c3.xlarge -> c4.xlarge
```

T2 medium instance receives 1 hour 100% CPU load burst upon creation, making it also suitable for compilation and errands.

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml.erb
+++ b/spec/fixtures/aws/cf-manifest.yml.erb
@@ -1145,7 +1145,7 @@ releases:
 resource_pools:
 - cloud_properties:
     availability_zone: zone_1
-    instance_type: c3.large
+    instance_type: t2.small
     ephemeral_disk:
       size: 10_240
       type: gp2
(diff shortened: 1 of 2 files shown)


zaksoup:
Hi @mtekel, We frequently evaluate which instances we choose as the default and in our exploration we discovered that the network performance of the T2 instance types is severely degraded. While you can certainly overwrite these instance types in your own manifests we have found that in our staging and production deployments the T2 instances are unable to sustain medium to heavy load and therefor cannot be recommended as defaults.

Thanks,
@zaksoup and @medvedzver, CF Release Integration

zaksoup:
I'd like to add that you can view the progress of these stories for looking into cheaper instances types on our tracker: https://www.pivotaltracker.com/story/show/103821782

mtekel:
Hi Zak,

it's not quite obvious that 'small' instance type should be able to sustain heavy load. Most of the small/medium jobs just spend time sitting idle. In case you need high performance, you can deploy multiple instances, or override them. Or perhaps change instance type of the job that requires high performance to larger (e.g. from medium_z1 to large_z1). c3.large for everything is overkill...

PR #800, opened 2015-09-30
Modify DEA configuration for higher performance

rexxar-liang:
- nf_conntrack_max is set to 65536 by default, we found linux kernel
  drop tcp handshake request with the message "nf_conntrack: table
  full, dropping packet" by using command "dmesg"
- in order to support higher performance, we set nf_conntrack_max
  to a larger number

Signed-off-by: Yun Lv <lvyun@huawei.com>


Amit-PivotalLabs:
@rexxar-liang How much memory do you have allocated to your DEAs?

Amit-PivotalLabs:
Just an update @rexxar-liang , we plan to pull this PR in.

Thanks,
Amit

PR #810, opened 2015-10-14
Update manifest generation spec stub for AWS to be more realistic

Amit-PivotalLabs:
This is coupled to cloudfoundry/loggregator#89.


Amit-PivotalLabs:
The original changes in this PR are in as of https://github.com/cloudfoundry/cf-release/commit/cc579c34ea587bf0d3110a6d2891d11602bfba25

PR #812, opened 2015-10-15
Cleanup UAA job spec and templates

sykesm:
This removes default references to run.pivotal.io in the UAA specs and enables consumers to disable the signup and password reset links from the default login page on the UAA.

The `links.home`, `links.network`, and `links.signup-network` do not appear to be used in the UAA so they were removed entirely. If they are used, the Pivotal Network related links should likely be extracted and separated from the open source release.

### DIFF
--- a/jobs/uaa/spec
+++ b/jobs/uaa/spec
@@ -314,10 +314,10 @@ properties:
     description: "Array of the router IPs acting as the first group of HTTP/TCP backends. These will be added to the proxy_ips_regex as exact matches. When using spiff, these will be router_z1 and router_z2 static IPs from cf-jobs.yml"
     default: []
   login.brand:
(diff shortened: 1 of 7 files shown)


sreetummidi:
@sykesm Looks good. Will get this merged in

Amit-PivotalLabs:
@mbhave @sreetummidi Can you confirm whether or not the `login.links` properties are being used any more?  I see the template does `login.links.marshal_dump` (so you don't directly see `login.links.home` referenced, for example), but does the UAA code in cf-release actually use them?

sreetummidi:
login.links is being used.
login.tiles is no longer in use

mbhave:
Actually login.tiles is used. @sykesm is right about login.links.home, login.links.network and login.links.signup-network not being used. We'll merge this in. Thanks.

PR #817, opened 2015-10-18
possibility to add custom SCIM groups to UAA automatically

MatthiasWinzeler:
We'd like to add certain SCIM groups to the UAA automatically (without the manual task of creating it with `uaac`). 

The cf-release UAA job allows specifying `authorities` - groups that are automatically created in the UAA db, but also always assigned to every user by default. We'd like to specify a set of SCIM groups in the config, that are automatically added to the UAA db, but not assigned to every user by default.

I saw the UAA config already allows specifying such groups in its [config](https://github.com/cloudfoundry/uaa/blob/bbea63986bbf2de9c42f231668e344a4a321184c/uaa/src/main/webapp/WEB-INF/spring/scim-endpoints.xml#L275), but it's not yet possible in the cf-release templates.

This PR adds the possibility to specify these groups as job properties.

### DIFF
--- a/jobs/uaa/templates/uaa.yml.erb
+++ b/jobs/uaa/templates/uaa.yml.erb
@@ -167,8 +167,8 @@ scim:
     - <%= group %><% end %>
 <% end %>
 
-<% if properties.uaa.scim && properties.uaa.scim.groups then %>
-  groups: <%= properties.uaa.scim.groups %>
+<% if_p('uaa.scim.groups') do |groups| %>
+  groups: <%= groups %>
 <% end %>
 
 <% if properties.uaa && properties.uaa.zones && properties.uaa.zones.internal &&  properties.uaa.zones.internal.hostnames then %>
--- a/jobs/uaa/templates/uaa.yml.erb
+++ b/jobs/uaa/templates/uaa.yml.erb
@@ -167,6 +167,10 @@ scim:
     - <%= group %><% end %>
 <% end %>
 
+<% if properties.uaa.scim && properties.uaa.scim.groups then %>
+  groups: <%= properties.uaa.scim.groups %>
+<% end %>
+
 <% if properties.uaa && properties.uaa.zones && properties.uaa.zones.internal &&  properties.uaa.zones.internal.hostnames then %>
 zones:
   internal:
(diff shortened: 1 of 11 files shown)


jobs/uaa/templates/uaa.yml.erb
```diff
@@ -167,6 +167,10 @@ scim:
     - <%= group %><% end %>
 <% end %>
 
+<% if properties.uaa.scim && properties.uaa.scim.groups then %>
+  groups: <%= properties.uaa.scim.groups %>
+<% end %>
+
```
Amit-PivotalLabs: This should use more conventional bosh helpers (unfortunately, the rest of the uaa templates are not very conventional so it's hard to discover this pattern).

```
<% if_p('uaa.scim.groups') do |groups| %>
  groups: <%= groups %>
<% end %>
```

Amit-PivotalLabs:
I'm not super familiar with UAA; @sreetummidi, @mbhave does this property change make sense?  Is `groups` a configurable thing UAA consumes, as a comma-separated list?

MatthiasWinzeler:
Hi Amit; 
`groups` is indeed a configurable comma-separated list that the UAA consumes, see here a test example:
https://github.com/cloudfoundry/uaa/blob/bbea63986bbf2de9c42f231668e344a4a321184c/uaa/src/test/resources/test/profiles/mysql/uaa.yml#L16

jobs/uaa/templates/uaa.yml.erb
MatthiasWinzeler: Thanks for the hint! I'm aware of that helper, but wanted the exact similar behaviour as the `external_groups` part which was in the "old" style. Can change it if you like?

mbhave:
@MatthiasWinzeler thanks for the PR. This is a configurable property that the UAA consumes.

jobs/uaa/templates/uaa.yml.erb
Amit-PivotalLabs: Yes please!

MatthiasWinzeler: done!

sreetummidi:
@MatthiasWinzeler The UAA team will take care of merging this PR in.  You can view the status on tracker. https://www.pivotaltracker.com/story/show/105949634

PR #822, opened 2015-10-23
Allow additional configuration of etcd

nickwei84:
We want to use the `election_timeout_in_milliseconds` parameter for etcd, but the templates don't support this yet. This change should allow use of any config parameters.

### DIFF
--- a/templates/cf-jobs.yml
+++ b/templates/cf-jobs.yml
@@ -641,6 +641,7 @@ properties:
     prof_port: 0
 
   etcd:
+    <<: (( merge ))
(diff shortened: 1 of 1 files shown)


Amit-PivotalLabs:
Hi @nickwei84 

We generally don't want to support arbitrary parameter configuration in the spiff templates.  The spiff flow would ideally be simple, explicity, and easy to understand.  This is unfortunately not currently the case, but we don't want to exacerbate the problem.

In this case, a preferable option would be:

```
election_timeout_in_milliseconds: (( merge || nil ))
```

Similar to the DEA configuration below in that same file.

nickwei84:
Cool I will look into the alternative option. Thanks!

PR #831, opened 2015-10-27
fix test timeout config bug

aocole:
Currently, the documented `cf_push_timeout`, `long_curl_timeout`, and `broker_start_timeout` options for the acceptance tests are ignored, and the default timeout is used instead. Looks like a copy-paste error.


robdimsdale:
Yes they looked like copy-paste errors. Thanks for submitting - merged.

PR #833, opened 2015-10-28
Added port: variable to router: properties

axelaris:
This patch adds ability to import router's `port` and `status.port` variables from stub file.
It needs when you move router template somewhere, where default ports are busy.


Amit-PivotalLabs:
Hi @axelaris 

Thanks for this PR.  Seems reasonable.

This change will cause unit tests to fail.  Please run `./scripts/setup-git-hooks` (see Step 9 of [general workflow in Contributing documentation](https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md#general-workflow)).  It will ensure there are no unused blobs as well as run the manifest-generation unit tests.

axelaris:
Fixed. Could you please check once again?

Amit-PivotalLabs:
Hi @axelaris 

Please leave the template values as `nil`, not `80` or `8080`.  It becomes confusing when the value can come from (a) any number of stubs, (b) the templates, (c) defaults in the job specs.  By putting `nil`, it ensures that if the value is not specified in the stub(s), then the value gets delegated to the job spec default.  If you run the tests, you'll see that the tests are failing because the fixtures are missing things like `port: null` in the appropriate places, so you can just add those to the fixtures.

Please also squash the commits into 1, so there don't end up being any failing commits on the develop branch.

Speaking of which, please submit all PRs against the develop branch.  You may want to review the [CONTRIBUTING.md](https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md) document, which talks about using the develop branch, running tests, signing the CLA, etc.

... Speaking of the CLA, @cfdreddbot's [comment above](https://github.com/cloudfoundry/cf-release/pull/833#issuecomment-151827022) seems to suggest that it doesn't have your CLA on record, do you know if you (or perhaps your organization if you're part of one) has already signed one?

Finally, regarding our discussion in cloudfoundry/cf-release#834, yes, please go ahead and add that to this PR, making sure to squash everything down to one commit (and rebasing on top of develop) before re-submitting.  I will close out the other PR.

Thanks,
Amit

axelaris:
Closing in favour of adding the desired change to https://github.com/cloudfoundry/cf-release/pull/837

axelaris:
@Amit-PivotalLabs, thank you for comments.
I've gave up trying to correctly squash all commits together, then decided to start all from scratch here: https://github.com/cloudfoundry/cf-release/pull/837. 

As for CLA, I'm pretty sure that my organisation (@Altoros) have it signed. Here is my colleague profile, as I can see, he has a lot to patches to *cf-release*: @allomov

allomov:
Actually, I am not sure about the way I signed CLA - it was 2 years ago. @Altoros has corporate CLA, but still I believe you need to reach someone from @Altoros management staff with this question.

PR #834, opened 2015-10-28
Changed default port from 17001 to 18001

axelaris:
17001 interferes with hardcoded value in `hm9000_analyzer`:
https://github.com/axelaris/cf-release/blob/master/jobs/hm9000/templates/hm9000_analyzer_ctl#L25

### DIFF
--- a/jobs/gorouter/spec
+++ b/jobs/gorouter/spec
@@ -36,7 +36,7 @@ properties:
     default: -1
   router.debug_addr:
     description: "Address at which to serve debug info"
-    default: "0.0.0.0:17001"
+    default: "0.0.0.0:18001"
   router.secure_cookies:
     description: "Set secure flag on http cookies"
     default: false


Amit-PivotalLabs:
Hi @axelaris 

One generally does not colocate the gorouter with the hm9000.  If you are doing this, you can set the `router.debug_addr` property to some value other than the default in your manifest.

axelaris:
Hi @Amit-PivotalLabs 
thank you for a reply.
Yes, I can add the `router.debug_addr` property to my stub file, but in fact, it will not be imported into the manifest by standard `generate_deployment_manifest` script:
https://github.com/cloudfoundry/cf-release/blob/master/templates/cf-jobs.yml#L704-L719

So, basically I have two options here: patch `cf-jobs.yml` to add a merge statement for `router.debug_addr` or patch the `spec` file to change the default value.
I believe changing default is better ;-)

Amit-PivotalLabs:
Hey @axelaris 

I'd recommend *not* changing the default.  It's a backwards incompatible change for people expecting the debug address to be on the 17001 port, not a huge deal, but I don't see a strong reason to break that compatibility.  Also, router and hm9k are independent, and developers working on one should not have to be worrying about what ports they put their services on and if it might clash with some service of some totally separate project.  BOSH gives you the flexibility to override the port in the case that you're doing something unconventional like colocating everything, so I'd prefer not to change the default port.

If you'd like to submit a PR to make that property spiff-able, we can probably merge it.

Cheers,
Amit

axelaris:
Hi @Amit-PivotalLabs,
would do you like me to add `router.debug_addr` to https://github.com/cloudfoundry/cf-release/pull/833 or I have to create a separate PR?

Amit-PivotalLabs:
Yes please.

Closing in favour of adding the desired change to cloudfoundry/cf-release#833.

PR #835, opened 2015-10-30
Add route registrars

axelaris:
This PR is like https://github.com/cloudfoundry/cf-release/pull/833 and https://github.com/cloudfoundry/cf-release/pull/834 and mainly related to moving CF components between their jobs.
Here I'm adding route_registrar's properties to jobs `etcd_z1`, `nfs_z1` and `route_z1` to make it possible to move `route_registrar` template there.

Unit tests also have been fixed.
I'm hope you will like this PR. 
Thanks.

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml.erb
+++ b/spec/fixtures/aws/cf-manifest.yml.erb
@@ -123,6 +123,8 @@ jobs:
   properties:
     metron_agent:
       zone: z1
+    route_registrar:
+      routes: null
   resource_pool: medium_z1
   templates:
   - name: etcd
@@ -175,6 +177,8 @@ jobs:
   properties:
     metron_agent:
       zone: z1
+    route_registrar:
+      routes: null
   persistent_disk: 102400
   resource_pool: medium_z1
   templates:
@@ -596,6 +600,8 @@ jobs:
           gorouter: {}
     metron_agent:
       zone: z1
+    route_registrar:
+      routes: null
   resource_pool: router_z1
   templates:
   - name: gorouter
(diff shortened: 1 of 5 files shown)


axelaris:
As for CLA, since I know my organisation (Altoros) have it signed already.

Amit-PivotalLabs:
Hey @axelaris 

Sorry, I didn't understand the purpose of this PR.  When do NFS, etcd, or router ever need to broadcast routes?

axelaris:
Hi @Amit-PivotalLabs,
this is about moving CF components between their jobs.
Let's imagine I want to prepare more compact deployment than standard one (6-8 VMs instead of 16). In order to move UAA, API or HM9000 components out from their original jobs (to share jobs with `etcd_z1`,  `nfs_z1` and `route_z1`) I have to make it possible to register routes from these jobs. But I can't without this patch.

Amit-PivotalLabs:
I fully understand your goal of wanting to do a CF deployment with a smaller VM footprint, however this solution seems very specific to the particular jobs you want to colocate with other jobs, and generally makes the interface confusing (it's not clear to anyone else why etcd needs to broadcast routes).

The spiff templates are not designed to support arbitrary use cases (for instance, it's limited to 2 AZs).  We don't plan on supporting this particular configuration in the spiff templates at this time.  You just need to make sure that in your deployment manifest, the UAA route registration configuration is present on any jobs that have the UAA jobs/templates colocated, and likewise for CC, HM9k, and Doppler.

We try to encourage people to write simple post-processing scripts against the spiff-generated manifests if they need to do something specific to their use case, rather than trying to support the hundreds/thousands of real-world tweaks people end up needing.  In other words, they are supposed to make deploying CF way easier than generating manifests by hand, but not to solve 100% of all use cases.  Hopefully this rationale makes sense.

Thanks,
Amit

PR #852, opened 2015-11-13
Error when BOSH_RELEASES_DIR doesn't exist

SocalNick:
Previously filed bug asking BOSH_RELEASES_DIR environment variable to be
required instead of silently using ~/workspace. Requiring the
environment variable might be a bit heavy handed from a UX standpoint,
how about we error if the BOSH_RELEASES_DIR doesn't exist? This
preserves same behavior when user clones to ~/workspace, but exits 1 if
the directory doesn't exist, e.g. in a concourse pipeline.

### DIFF
--- a/scripts/generate-bosh-lite-dev-manifest
+++ b/scripts/generate-bosh-lite-dev-manifest
@@ -3,6 +3,11 @@
 export BOSH_USE_BUNDLER=true
 
 BOSH_RELEASES_DIR=${BOSH_RELEASES_DIR:-~/workspace}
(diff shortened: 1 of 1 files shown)


Amit-PivotalLabs:
Hey @SocalNick 

We're actually going to move that stub from diego-release into cf-release, and always merge it in for the BOSH-Lite case, and just get rid of all this BOSH_RELEASES_DIR silliness.

See: https://www.pivotaltracker.com/story/show/108282040

Closing this PR in lieu of that work.

PR #858, opened 2015-11-23
Fix CLA URLs in contributing guide

ches:
It appears that the PDF documents have moved.

I'm not sure if there's a `[ci skip]` type of commit message annotation for Concourse, if there is and it should be used I can update this commit.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -19,8 +19,8 @@ If you want to make changes to the `cf-release` itself, read on.
 Follow these steps to make a contribution to any of our open source repositories:
 
 1. Ensure that you have completed our CLA Agreement for
-  [individuals](http://www.cloudfoundry.org/individualcontribution.pdf) or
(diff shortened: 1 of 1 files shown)


ches:
I have submitted my own signed individual CLA by email, FYI.

Amit-PivotalLabs:
Hi @ches 

We ask PRs to be submitted against the `develop` branch, not `master`.  The CONTRIBUTING.md doc has already been updated on `develop` to point to the new site that talks about the CLA.

Thanks,
Amit

ches:
@Amit-PivotalLabs In that case, it would be helpful to consider setting `develop` as the default branch in the GitHub project administration. It would then a.) be the default tree when browsing on the web and the default branch for new clones, so a mistake like mine would be avoided; and b.) would be the default branch for pull requests.

Amit-PivotalLabs:
@ches the default branch is set to `master` because that is the stable branch that users of `cf-release` should use.  The `CONTRIBUTING.md` document explains using `develop` for contributions.

PR #859, opened 2015-11-23
Allow additional configuration for Postgres

fraenkel:
- additional_config can provide extra configuration tweaks to the postgres
  config


fraenkel:
With larger installations, additional configuration properties need to be adjusted. Rather than attempt to identify which ones are needed, we prefer to allow any to be specified or provide overrides to those that already are hard coded.

PR #860, opened 2015-11-23
Add license files to final packages.

aarondl:
The idea here is to deliver a bosh.io tarball with the licenses inside each
package.tgz file distributed. The tarball ships with source code not simple
binaries and therefore should probably be shipped with licenses alongside.

- CF Slack participants told me to go ahead and open these PRs.
- This work is semi-dependent on other bits in other submodules but I was
  instructed not to commit submodule pointers and instead do separate
  PRs and such the submodules/repos: uaa, loggregator, etcd-release,
  statsd-injector should also be merged/updated.

https://github.com/cloudfoundry-incubator/etcd-release/pull/8
https://github.com/cloudfoundry/statsd-injector/pull/1
https://github.com/cloudfoundry/uaa/pull/256
https://github.com/cloudfoundry/loggregator/pull/100

### DIFF
--- a/packages/cloud_controller_ng/spec
+++ b/packages/cloud_controller_ng/spec
@@ -15,3 +15,5 @@ files:
 - cloud_controller_ng/config/boot.rb
 - cloud_controller_ng/config/environment.rb
 - cloud_controller_ng/config/routes.rb
+- cloud_controller_ng/LICENSE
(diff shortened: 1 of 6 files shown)


robdimsdale:
Thanks!

Rob & @zaksoup

PR #862, opened 2015-12-02
add script to generate manifest from github

jim80net:
This script allows for manifest generation without a checkout of the repo.

It would be used from a wrapper script (an example is provided inline with the code)

Signed-off-by: MKB <mkb@starkandwayne.com>

### DIFF
--- a/scripts/generate_deployment_manifest_github
+++ b/scripts/generate_deployment_manifest_github
@@ -53,9 +53,4 @@ do
   full_path_to_templates+="$tmpdir/$i "
 done
 
-##
-# CF-LAMB is a symlink to a submodule
-SHA_OF_CF_LAMB=$(curl https://api.github.com/repos/cloudfoundry/cf-release/contents/src/loggregator?ref=${myref} 2>/dev/null | jq --raw-output '.sha')
-curl https://raw.githubusercontent.com/cloudfoundry/loggregator/${SHA_OF_CF_LAMB}/manifest-templates/cf-lamb.yml >${tmpdir}/cf-lamb.yml 2>/dev/null
-
 spiff merge $full_path_to_templates "$@"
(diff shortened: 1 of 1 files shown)


amitkgupta:
As of the next final release, there will be no more templates in submodules (see current release-candidate branch, templates directory, for instance).  Given that fact, does this script still serve much of a purpose?

jim80net:
updated PR to reflect that, thanks @amitkgupta. The script would help CloudOps out (and other operators, I imagine), by allowing an operator to generate a manifest without having to have a checkout of cf-release, which would save a bunch of time and downloading.

Amit-PivotalLabs:
@jim80net 

I still see several issues with this PR (some small, some not so small):

1. Comments mention `pivotal-cloudops`
2. There is no test coverage for the script
3. It seems impossible to keep up to date: if I change the names of the templates, there's no way to have a passing test for this script that reflects the current ref until I push the changes up to GitHub first, so I would have to push before the test passes, defeating the purpose
4. There is no cf-lamb.yml; the change that makes the loggregator submodule unnecessary also removes the `cf-lamb.yml` template and includes the relevant templating in the other template files (this is perhaps stark proof of point 3)
5. It leaks the implementation of `generate_deployment_manifest`, basically any logic in that script needs to be duplicated for this script to stay up to date.
6. I'm not sure the point of having this in cf-release itself.

What if you just did:

```
$ git clone --depth=1 --branch ${desired_branch} https://github.com/cloudfoundry/cf-release /tmp/hi-mom
$ cd /tmp/hi-mom && ./scripts/generate_deployment_manifest ${iaas} ${OLD_PWD}/stubs/my-stub.yml
```

As long as you only every need branches or tags (not arbitrary refs) it's pretty fast and much more guaranteed to be up to date.

jim80net:
You're right for all of those points. The workaround provided doesn't work for us, because we do sometimes want arbitrary refs. However, we found a way to do this without injecting things into cf-release: https://gist.github.com/jim80net/9f0a9ca9679f664cca08

Thanks!

Amit-PivotalLabs:
Cool, glad you still found something that worked.

PR #865, opened 2015-12-14
Graceful drain

CAFxX:
Enabling bits for https://github.com/cloudfoundry/gorouter/pull/109

Expose drain_wait with a default value of 0 seconds (i.e. the current behavior: do not wait before stopping the listener)


CAFxX:
sorry, wrong target branch

PR #875, opened 2016-01-06
Use new managed NAT gateway

Infra-Red:
This will simplify minimalistic deployment of Cloud Foundry. New managed NAT gateway can be created and configured with a couple of clicks, instead of launching EC2 NAT instance.

### DIFF
--- a/example_manifests/README.md
+++ b/example_manifests/README.md
@@ -3,7 +3,7 @@
 The `minimal-aws.yml` is an example of a minimalistic deployment of Cloud
 Foundry, including all crucial components for its basic funcionality. it allows
 you to deploy Cloud Foundry for educational purposes, so you can poke around and
-break things. 
+break things.
 
 *IMPORTANT*: This is not meant to be used for a production level deployment as
 it doesn't include features such as high availability and security.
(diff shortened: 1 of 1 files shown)


Amit-PivotalLabs:
Thanks @Infra-Red!

I've never walked through these instructions myself, so I'll try it out and make sure the new NAT gateway instructions work with the rest of the existing instructions.  Might take me a little while to get to this.  In the mean time, we'll also wait for your CLA.

Amit-PivotalLabs:
@Infra-Red have you noticed any issues with your NAT Gateway? Some colleagues of mine have reported flakiness, as have some others in general as well: https://forums.aws.amazon.com/thread.jspa?threadID=222925

Infra-Red:
@Amit-PivotalLabs I have not seen such problems and my current installation works as expected, but I do not exclude the possibility of problems with the new service... Also i made some simple tests and didn't faced this issue.

Amit-PivotalLabs:
Got this email from a colleague:

> There's a known issue with the rubygems CDN (Fastly) and the NAT Gateway. Googling with this new information returned the following thread: https://forums.aws.amazon.com/thread.jspa?messageID=696009

> There's a good description of the problem in the response. AWS support has not given a timeline for a fix, so you'll want to avoid the NAT Gateway service for the time being.

Seems like this will make staging ruby apps flaky.

PR #878, opened 2016-01-12
make hm9000 port configurable

ronakbanka:


fraenkel:
Thank you for your pull request. It just so happens that we were in the process of making a similar change. I will document the commit in this PR for your awareness.

fraenkel:
See commit d189dc6af201156816585ccea0d64be87829d2d3

PR #879, opened 2016-01-12
default port for nats

ronakbanka:


Amit-PivotalLabs:
Hi @ronakbanka 

It would be great if it were possible to specify the nats port default in the job spec like this, and have all clients of nats automatically discover this configuration.  This kind of discovery will be possible with the upcoming [links feature in BOSH](https://github.com/cloudfoundry/bosh-notes/blob/master/links.md).

Unfortunately, for now, all other clients still need to be configured with the nats port value in the manifest, so there isn't much benefit in setting the default here in the job spec for the nats server.

ronakbanka:
@Amit-PivotalLabs  Agree it will be useful with bosh links feature.

For now nats port is configured from manifest but as nats port is kind of standard with most of the deployments , would't it be good to add this default just for nats server itself. 

No Hard requirements but just for convenience.

Amit-PivotalLabs:
I think this default buys you very little, and can only lead to problematic situations, e.g. if the default value changes or the operator doesn't configure it and doesn't realize anything is wrong because the nats job template will not blow up if not configured.  I'd rather force the user to explicitly configure it everywhere (until we start using BOSH Links) than to make it easy to make accidental mistakes.  Also, I believe the spiff templates depend on it being explicitly set for the nats server.

Will close this PR for now.

Best,
Amit

PR #880, opened 2016-01-14
Check for spiff installation

tomoe:
Spiff command is used in the generate_deployment_manifest
script, but its installation was not checked.
Now the script checks if spiff is installed and if it's not
installed, it exits with a friendly message guiding to
the installation doc.

### DIFF
--- a/scripts/generate_deployment_manifest
+++ b/scripts/generate_deployment_manifest
@@ -8,6 +8,12 @@ fi
 
 set -u
 
+# Check for spiff installation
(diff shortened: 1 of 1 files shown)


robdimsdale:
Looks good - we will merge shortly.

Thanks,
Rob && @zaksoup

robdimsdale:
Merged, thanks

tomoe:
NP. Thanks for the quick turnaround.

PR #881, opened 2016-01-14
fix example_manifests README

mirkoperillo:
In example_manifests/minimal-aws.yml the bosh stemcell to use is `bosh-aws-xen-hvm-ubuntu-trusty-go_agent` but README instructions report to download/upload `bosh-aws-xen-ubuntu-trusty-go_agent`

### DIFF
--- a/example_manifests/README.md
+++ b/example_manifests/README.md
@@ -138,14 +138,14 @@ Download a copy of the latest bosh stemcell.
 
 ```sh
 bosh public stemcells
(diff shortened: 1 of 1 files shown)


robdimsdale:
@mirkoperillo thanks for this. Good catch. We would like to pull this in but it needs to be targeted against the `develop` branch, not master.

Also, we'll need you to sign a CLA - see the comment above from cfdreddbot.

Once this two things are addressed in a separate PR we will merge.

Thanks,
Rob & @zaksoup

PR #883, opened 2016-01-16
cats errand is configurable for new diego tests

cwlbraa:
this is blocked on https://www.pivotaltracker.com/story/show/111813688

[#111125958]

Signed-off-by: Amelia Downs <adowns@pivotal.io>


Amit-PivotalLabs:
Following the discussion on [Slack](https://cloudfoundry.slack.com/archives/release-integration/p1453416656001845), we are going to fix this PR by removing the `skip_diego_unsupported` property, and update which `NO_{DEA,DIEGO}_SUPPORT` tests we skip based on the value of the `backend` property.

/cc @zaksoup @cwlbraa @ematpl @dsabeti

PR #885, opened 2016-01-20
Updated shebang lines

SamuelMarks:
`generate-bosh-lite-dev-manifest` wouldn't run for me, changing the shebang line worked. So I wrote this little snippet to change all the shebangs to `#!/usr/bin/env bash`:
```
find -type f ! -name "*.*" -exec sed -i '1{/^#!\/[^ ]*\/\(ba\|\)sh\( *\)/s||#!/usr/bin/env bash\2|}' {} \;
```

Recommend doing similar throughout your repositories.

### DIFF
--- a/git_hooks/pre-commit
+++ b/git_hooks/pre-commit
@@ -1,2 +1,2 @@
-#!/bin/sh
+#!/usr/bin/env bash
 ./scripts/test
--- a/jobs/debian_nfs_server/templates/rpc_mountd_ctl
+++ b/jobs/debian_nfs_server/templates/rpc_mountd_ctl
@@ -1,4 +1,4 @@
-#!/bin/bash
+#!/usr/bin/env bash
 
 RUN_DIR=/var/vcap/sys/run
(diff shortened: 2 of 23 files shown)


fhanik:
`/bin/sh` is standard on most Unix systems, not just Linux. You may want to change(fix) your system instead of changing the scripts.

Amit-PivotalLabs:
Hey @SamuelMarks 

Thanks for the PR.  Couple things:

- anywhere that was using `/bin/sh` should continue to use `sh` instead of `bash`.  You could use `/usr/bin/env sh` in these cases.
- scripts under the `jobs` directory run inside of pre-baked OS images, generally as part of a Cloud Foundry deployment, not on developer machines.  These should not be changed.

Also, curious to understand your use case and your dev env, any reason your `bash` isn't in the standard location?

SamuelMarks:
Turns out it was a line endings issues. Nevermind, I'm sure this snippet will be useful at some other point

PR #891, opened 2016-01-21
Allow customization of uaa.client property

cholick:
uaa's spec also has a `client` property that we need to customize in what we merge in:
https://github.com/cloudfoundry/uaa-release/blob/develop/jobs/uaa/spec#L229

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml
+++ b/spec/fixtures/aws/cf-manifest.yml
@@ -1208,6 +1208,7 @@ properties:
     catalina_opts: -Xmx768m -XX:MaxPermSize=256m
     cc:
       client_secret: CC_CLIENT_SECRET
+    client: null
     clients:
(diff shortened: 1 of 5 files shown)


Amit-PivotalLabs:
Hey @cholick, do you have the git hooks for the cf-release repo installed? (Step 9 in the [contributing docs](https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md#general-workflow)). The hooks will cause manifest generation tests to run when commiting. I haven't checked, but I suspect the manifest generation tests break with this change, but should be a simple fix.

cholick:
I don't; missed that. Will install them and fix up any errors.

cholick:
Should be good now.

cholick:
We came at this differently & so no longer need to customize this property.

PR #892, opened 2016-01-22
update exports template 

dliebreich:
- nfs_server.allow_from can now be an array of arrays

discussed with greg, dan, and dieu

reviewed by @davewalter and @sks


simonleung8:
Looks good. Merged.

PR #895, opened 2016-01-27
Allows overriding of bucket names.

mavenraven:
Allows overriding of bucket names. See http://shlomoswidler.com/2009/08/amazon-s3-gotcha-using-virtual-host.html

### DIFF
--- a/templates/cf-properties.yml
+++ b/templates/cf-properties.yml
@@ -104,18 +104,18 @@ properties:
     broker_client_max_async_poll_duration_minutes: (( merge || 10080 ))
 
     resource_pool:
-      resource_directory_key: (( domain "-cc-resources" ))
+      resource_directory_key: (( merge || domain "-cc-resources" ))
       fog_connection: ~
       cdn: ~
 
     packages:
-      app_package_directory_key: (( domain "-cc-packages" ))
+      app_package_directory_key: (( merge || domain "-cc-packages" ))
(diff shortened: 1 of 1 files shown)


mavenraven:
Also, I want this because I want to use a guid like string instead of domain name for my buckets.

Amit-PivotalLabs:
Hey @mavenraven 

Have you tried these changes out, and seen everything work? Deploy goes through, CATS pass, etc?

mavenraven:
Nope. I'd assume it wouldn't break any existing manifest generation since the default behavior should remain the same, but I don't have a good enough computer / account to run the tests at home.

Amit-PivotalLabs:
Have you actually used these changes? Have you deployed to AWS leveraging these changes? If so, could you run the CATS errand (if not able to run CATS from your laptop) to see if everything passes?

Amit-PivotalLabs:
Hey @mavenraven 

Let me know if you've had a chance to validate these changes in real life.  If you don't think you'll get a chance to exercise these changes, let us know so we can close out the PR.

Cheers,
Amit

mavenraven:
@Amit-PivotalLabs I do plan on validating this, but it probably won't be for a couple of weeks.

Amit-PivotalLabs:
Cool.  I'll close out the PR at this time, feel free to open a new one when
you need it.

PR #915, opened 2016-02-23
HCF-516: Make the name of the CCNG hostname configurable

ericpromislow:
This allows HCF to not depend on consul to do DNS resolution. Also it's the
only hardwired hostname in cf. Works with ccng PR https://github.com/cloudfoundry/cloud_controller_ng/pull/544

### DIFF
--- a/jobs/cloud_controller_clock/spec
+++ b/jobs/cloud_controller_clock/spec
@@ -67,6 +67,9 @@ properties:
   cc.external_port:
     description: "External Cloud Controller port"
     default: 9022
+  cc.internal_service_hostname:
+    description: "Internal hostname used to resolve the address of the Cloud Controller"
+    default: "cloud-controller-ng.service.cf.internal"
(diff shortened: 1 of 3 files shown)


Amit-PivotalLabs:
Hi @ericpromislow 

How exactly do you plan on using this?  I believe this internal hostname is handed over to Diego to tell it where to get app bits, buildpacks, and blobs when CC requests a staging or app running.  Do you plan on running some other DNS that the Diego Cells can use to take the given internal hostname and have it resolve to all the CCs?

ericpromislow:
Yes, we're using a different DNS to resolve the location of the CCs, and it needs a different value.

Amit-PivotalLabs:
Thanks @ericpromislow,

I'll ask the Cloud Controller team if they can take a look.

Best,
Amit

zrob:
merged on develop

PR #921, opened 2016-03-03
Set deployment name to cf

holgero:
The deployment name for cloud foundry should be cf by default and not include the infrastructure name.
Other cf-infrastructure-* stubs also do not include the name of their infrastructure into the deployment name.

### DIFF
--- a/templates/cf-infrastructure-warden.yml
+++ b/templates/cf-infrastructure-warden.yml
@@ -1,5 +1,5 @@
 meta:
-  environment: cf-warden
+  environment: cf
 
   stemcell: (( merge || .meta.default_stemcell ))
 


Amit-PivotalLabs:
Hey @holgero 

Any particular reason you're interested in this change, or just for consistency?  I think it's a fine change, just curious.  It will break people who already have `cf-warden` deployed, they'll have to know to delete their old deployment, but this isn't a big deal for BOSH-Lite environments.

holgero:
Hey @Amit-PivotalLabs,
basically it's for consistency. I stumbled over this when I was creating some dashboards in grafana. The time series name contains the deployment name and so I had to work around this difference. I wanted to be able to use the same dashboard definitions on my bosh-lite installation as on the productive systems on openstack. One idea was to patch the dashboard definitions when installing them, but I think the better solution is to just have the same deployment name 'cf' on all infrastructures.

Amit-PivotalLabs:
This is actually not consistent with the other manifest generation templates.  The deployment name is determined by the `meta.environment` value, none of them default to "cf".  You can set `meta.environment` in your stub for BOSH-Lite just as you can for any other IaaS.

PR #922, opened 2016-03-03
Allow specifying a reserved set of private domains

fraenkel:
This goes with
https://github.com/cloudfoundry/cloud_controller_ng/pull/554

The corresponding CC changes will need to be included.


Amit-PivotalLabs:
@SocalNick @utako looks like these changes were lost in the move to capi-release. /cc @dsabeti @fraenkel

utako:
Hi @fraenkel,

I've re-added those changes into the capi-release specs for the three CC jobs.

Thanks,
@utako

PR #935, opened 2016-03-19
Consolidate manifest templates

Amit-PivotalLabs:


spec/fixtures/bosh-lite/cf-manifest.yml
```diff
@@ -7,8 +7,8 @@ compilation:
 director_uuid: DIRECTOR_UUID
 jobs:
 - default_networks:
-    - name: cf1
-      static_ips: null
```
APShirley: This is valid yaml, but we (the team prefer arrays to be directly under the parent key rather than tabbed in)

spec/fixtures/bosh-lite/cf-manifest.yml
```diff
@@ -32,8 +32,8 @@ jobs:
     max_in_flight: 1
     serial: true
 - default_networks:
-    - name: cf2
-      static_ips: null
```
APShirley: See above

APShirley:
Looks good.

PR #944, opened 2016-04-01
Add setting for cc.core_file_pattern

alexjh:
Add default value for cc.core_file_pattern added in https://github.com/cloudfoundry/capi-release/pull/3 / https://github.com/cloudfoundry/cloud_controller_ng/pull/566.


alexjh:
Closing as this can be handled as a default in the CC spec.

PR #945, opened 2016-04-01
Needed `dropsonde_incoming_port: 3457` for BOSH deploy to succeed

cunnie:


amitkgupta:
We deploy this in CI all the time, and as far as I know we don't need to add this property in.  It's also already the default value in the job spec for `metron_agent`.  Can you give more info about your manifest and releases that failed without this property and succeeded with it?

youngm:
For a time DEA defined metron_agent.dropsonde_incoming_port without a default causing the need to add the property if bosh chose the dea spec property over the metron_agent spec property.  However, it appears that has been fixed in dea-hm-workspace.

https://github.com/cloudfoundry/dea-hm-workspace/blob/master/jobs/dea_next/spec#L159

youngm:
It is caused by the hm9000 job in dea-hm-worspace.  It declares `metron_agent.dropsonde_incoming_port` without a default.  I'm not sure how bosh chooses spec properties when multiple jobs define the same property but it appears that sometimes bosh chooses the hm9000 property over the metron_agent property causing the need to set this value.

https://github.com/cloudfoundry/dea-hm-workspace/blob/master/jobs/hm9000/spec#L66

fraenkel:
That will be fixed in the next bump.

Personally I think if a component requires another component to have the same value there should be no spec default. There is no way to share them other than specified in the manifest.

Amit-PivotalLabs:
@fraenkel I agree.  We'll merge this PR anyways since without BOSH Links, this shared config belongs in the manifest, not in job spec defaults.

PR #951, opened 2016-04-18
openstack: Adding static IP to jobs on z2 network

fbonelle:
This patch makes easier to enable jobs on z2 network in case of an Openstack deployment. It does not alter the default behaviour.


Amit-PivotalLabs:
Thanks @fbonelle this is great.

Do you mind making the PR against the `develop` branch?

fbonelle:
Sorry, I make the pull request against the right branch

PR #952, opened 2016-04-18
openstack: Add static IP to jobs on z2 network

fbonelle:
This patch makes easier to enable jobs on z2 network in case of an Openstack deployment. It does not alter the default behaviour.


dsabeti:
We verified that this won't change the behavior for single AZ deployments. We'll merge it. Thanks!

David && @robdimsdale 
CF Release Integration

PR #957, opened 2016-04-20
Change bosh-lite to deploy jobs in parallel

jpalermo:


Amit-PivotalLabs:
This will fail the manifest generation unit tests.  Also, PRs should be made against the `develop` branch.  Do you mind resubmitting (as a single squashed commit with code and tests updated)?

PR #958, opened 2016-04-21
Change bosh-lite to deploy jobs in parallel

jpalermo:


DennisDenuto:
We accidentally rebased this commit (instead of merging). But here's the commit: https://github.com/cloudfoundry/cf-release/commit/70ea123dcc411621ac1fe1f5b2e3ca19fbb5225e

PR #960, opened 2016-04-24
Remove unneeded router static IPs in AWS manifest template [#117196457]

Amit-PivotalLabs:
https://www.pivotaltracker.com/story/show/117196457


APShirley:
Merged, thanks!

Al & @robdimsdale

PR #963, opened 2016-05-11
Using --local to install bundler

### LINKED ISSUE #947: Can you add --local to only use local cache to install ruby 2.1.8 or 2.2.4?
Sometime the network to rubygems.org is not stable, especially in some countries. So when deploying cloud foundry, the tasks always fails in compiling ruby. 

Could you add --local to below two lines to only use local cache when compiling ruby?

[${BOSH_INSTALL_TARGET}/bin/gem install ruby-2.1.8/bundler-${BUNDLER_VERSION}.gem --no-ri --no-rdoc](
https://github.com/cloudfoundry/cf-release/blob/d58979fa9d54eeeebe71beca527e7745c4db237a/packages/ruby-2.1.8/packaging#L32)

[${BOSH_INSTALL_TARGET}/bin/gem install ruby-2.2.4/bundler-${BUNDLER_VERSION}.gem --no-ri --no-rdoc](https://github.com/cloudfoundry/cf-release/blob/d58979fa9d54eeeebe71beca527e7745c4db237a/packages/ruby-2.2.4/packaging#L31)

bingosummer:
Close #947


Amit-PivotalLabs:
Thanks for the PR @bingosummer 

Can you confirm that you were actually able to deploy Cloud Foundry using these changes?

Also, we ask that all contributors sign the CLA (see comment from @cfdreddbot above).  Please let us know when you have it signed.

Best,
Amit

bingosummer:
@Amit-PivotalLabs 

Yes. I am able to deploy Cloud Foundry with these changes.
If some tests or logs is needed from my side, please let me know.

For the CLA, a corporate one for Microsoft Corporation is signed at 09/22/2015.

Amit-PivotalLabs:
Thanks @bingosummer, no need for logs, your word is good enough!  Of course we verify everything locally before merging anyways.

Do you mind making your membership in the Microsoft GH organization public (see https://help.github.com/articles/publicizing-or-hiding-organization-membership/).  This way the bot can automatically detect your CLA status for this and all future PRs you submit to the cloudfoundry and cloudfoundry-incubator orgs.

bingosummer:
Done. Thanks @Amit-PivotalLabs

bingosummer:
Hi @dsabeti 
I noticed that the bundler binaries were removed when the PR was merged. If they are removed, can the PR work?
Thanks.

dsabeti:
Hey @bingosummer I think I understand your confusion. Allow me to explain as I understand it.

When you want to add a new binary dependency to a bosh release, the typical way to do it is to add it as a blob. You can see all the blobs in `config/blobs.yml`. This file references a list of binaries and tarballs stored in an s3 bucket. When you run `bosh create release`, the blobs get download and vendored into your release tarball (which is what gets uploaded to your bosh director when you run `bosh upload release`). So, if you look at https://github.com/cloudfoundry/cf-release/blob/develop/config/blobs.yml#L46-L49, you can see that bundler is included as a blob in the release.

The scripts you edited were the packaging scripts, which are run on compilation VMs after running `bosh deploy`. During this step, the packaging script is given all the blobs it needs -- in this case, `bundler-VERSION.gem`. Adding the `--local` flag to the packaging script ensures that the compilation VM will use the local gem instead of trying to fetch it from rubygems.

The binaries you included in the packaging directory actually have no effect on the packaging scripts -- the scripts are hard-coded to use the blobs. That why we removed them.

Have you been noticing problems with the packaging scripts? If so, we can definitely take a look.

**tl;dr** - Everything should still work fine because bundler is shipped as a blob.

bingosummer:
Thank @dsabeti  for explaining it.

In #947 , @APShirley mentioned that `It seems like the --local option expects the gems it installs to be present locally, which will not be true on the compilation VM because they are not vendored in as dependencies of the ruby package.`
Just to clarify that the gems will be local in the compilation VM. Right?

Another question about the release process. This PR has been labelled with `accepted` which means it will be released in `v238`. Right?

Thanks again for your patience.

Amit-PivotalLabs:
@bingosummer 

I'm actually a little confused by this PR now.  Since the `.gem` files is provided directly to the `install` command, and `bundler` has no non-development gem dependencies in its gemspec, it should never have any reason to go out to the Internet.  In the other related issue, it was mentioned that it "sometimes" fails.  Do you have any log output from failed compilations of this gem?  I'm actually not clear on what the `--local` flag accomplishes here.

bingosummer:
@Amit-PivotalLabs 
When compiling packages, it failed with the following output.
```
Started compiling packages > ruby-2.1.8/b5bf6af82bae947ad255e426001308acfc2244ee. Failed: Action Failed get_task: Task 542c2928-8bf3-4397-457b-450299e02aaf result: Compiling package ruby-2.1.8: Running packaging script: Command exited with 1;
<Truncated>
gcc version 4.8.4 (Ubuntu 4.8.4-2ubuntu1~14.04.1)
+ /var/vcap/packages/ruby-2.1.8/bin/gem install ruby-2.1.8/bundler-1.9.4.gem --no-ri --no-rdoc
ERROR:  While executing gem ... (Gem::RemoteFetcher::FetchError)
    Errno::ETIMEDOUT: Connection timed out - connect(2) for "api.rubygems.org" port 443 (https://api.rubygems.org/api/v1/dependencies?gems=bundler)
```

Amit-PivotalLabs:
Hmm, that's weird, because all that dependency information is already
encapsulated inside ruby-2.1.8/bundler-1.9.4.gem file.  I understand what
the PR is doing now, thanks.  Yes, it has been accepted and will be in the
next release (v238).

bingosummer:
Thanks Amit.

In my test, if `--local` is not specified, it sometimes fails due to the network issue in China. That's why we want to add `--local`. If that doesn't make sense, please let me know.

bingosummer:
@Amit-PivotalLabs 
Sorry to bother you again.
Could you please share a rough timeline about the `v238` release? I don't mean to push you guys. I totally understand you guys have your schedule. Just for a heads up.
Thank you.

Amit-PivotalLabs:
@bingosummer I plan on providing an update on the cf-dev and cf-bosh mailing lists tomorrow.

PR #965, opened 2016-05-16
Eliminate "MicroBOSH"; its use is deprecated

cunnie:
- The term "MicroBOSH" (single-VM deployment of a BOSH director) is deprecated. Its use is both unimportant and confusing.
- Instructions for deploying BOSH director refer to *bosh-init*; the *bosh micro* plugin is deprecated should not be referenced (note that *bosh micro* plugin has nothing to do with a microBOSH director, which is another reason to get away from the term "micro".

### DIFF
--- a/example_manifests/README.md
+++ b/example_manifests/README.md
@@ -10,10 +10,10 @@ it doesn't include features such as high availability and security.
 
 # Setup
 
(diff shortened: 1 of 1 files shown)


robdimsdale:
Merged, thanks @cunnie !

PR #966, opened 2016-05-20
Add minimum_candidate_stagers to manifest template

ScarletTanager:
[#119362989]

Signed-off-by: Dan Lavine <dlavine@us.ibm.com>

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml
+++ b/spec/fixtures/aws/cf-manifest.yml
@@ -917,6 +917,7 @@ properties:
         ca_cert: null
     bulk_api_password: BULK_API_PASSWORD
     client_max_body_size: 15M
+    minimum_candidate_stagers: 5
     dea_use_https: false
     db_encryption_key: CCDB_ENCRYPTION_KEY
     db_logging_level: debug2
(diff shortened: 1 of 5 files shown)


Amit-PivotalLabs:
Any reason this default of 5 isn't in the job spec rather than the manifest template @ScarletTanager @DanLavine?

DanLavine:
@Amit-PivotalLabs,

There is already a default of 5 in the job spec. However it seems kind of odd when looking at a manifest that was generated and this field having a nil value. We are just following the same scheme where the value of nil does not make much sense. (i.e. staging_disk_limit_mb)

/cc @ScarletTanager

Amit-PivotalLabs:
Sorry, I don't quite understand.  Is there some functional reason why the default value from the spec needs to be duplicated in the template (which is something we strongly try to avoid), or is it just that the manifest looks odd?

DanLavine:
@Amit-PivotalLabs 

It is just to make the generated manifest more readable. There is no technical reason that the values need to be duplicated. We are just being consistent in how these descriptive values are in both places.

Amit-PivotalLabs:
We strongly avoid duplicating default values in job specs and the manifest templates.  If we are consistently doing this in other places in the manifest templates, we should remove those as well.  How the manifest reads is less important at this time.  Thanks for the PR, but will close it out for now.

PR #973, opened 2016-06-07
Make possible to add new resource pools

v-boutes:
Currently we only can overload the existing resource pools. For enabling the use of OpenStack anti-affinity ServerGroup we had to add new resource pools and assign them by type of CF components. It was not possible because some resource pools were shared between differents types. So we add a merge instruction to add new custom resource pools.

### DIFF
--- a/templates/cf-infrastructure-openstack.yml
+++ b/templates/cf-infrastructure-openstack.yml
@@ -86,6 +86,7 @@ compilation:
   cloud_properties: (( merge || meta.resource_pools.medium.cloud_properties ))
 
 resource_pools:
+  - <<: (( merge ))
   - name: small_z1
     cloud_properties: (( merge || meta.resource_pools.small.cloud_properties ))
 
--- a/templates/cf.yml
+++ b/templates/cf.yml
@@ -1031,6 +1031,7 @@ update:
   serial: true
 
 resource_pools:
+  - <<: (( merge ))
   - name: small_z1
     network: cf1
     stemcell: (( meta.stemcell ))


Amit-PivotalLabs:
Hi @v-boutes 

Thanks for this PR.  I have a question: is there a reason to add this to spiff templating rather than adding your own resource pools to the manifest via some post-processing?

v-boutes:
Hi @Amit-PivotalLabs 

The pre-processing seems to be a best practice because it does not only allow the use of anti-affinity but it also opens the door to implement other cloud functions which will be useful to the community. We think there will probably be new use cases with the need to add resource pools.

Amit-PivotalLabs:
Sorry @v-boutes, I didn't quite understand that explanation.  Can you give an example of something that this PR allows you to do, that you couldn't do without spiff (either manually or with your own tooling) *after* spiff has generated a manifest?

v-boutes:
We have to provide "scheduler_hints" cloud property in resource pool and its value has to be different between CF component types. You can see it in our template defining resource_pools:

`
resource_pools:
- name: small_z1_consul
  cloud_properties:
    scheduler_hints:
      group: 68f27b47-5331-4737-a0f0-f4427f768417
- name: small_z1_api_worker
  cloud_properties:
    scheduler_hints:
      group: 3c56e409-3277-451c-81ac-26ae23b52941
...
Some output have been omitted
...
`

It’s currently not possible because some resource_pools are shared between different CF component types, for example "small_z1" with "consul" and "api_worker" (in cf-release/templates/cf.yml) :

`
...
Some output have been omitted
...
jobs:
  - name: consul_z1
    templates: (( merge || meta.consul_templates ))
    instances: 2
    persistent_disk: 1024
    resource_pool: small_z1
...
Some output have been omitted
...
  - name: api_worker_z1
    templates: (( merge || meta.api_worker_templates ))
    instances: 1
    resource_pool: small_z1
...
Some output have been omitted
...
`

It seems changing the manifest after generation could be more tedious.

Do you see something wrong to add this in template that i can't see ?

Amit-PivotalLabs:
You simply need to put something like:

```
...
jobs:
  - name: consul_z1
    resource_pool: small_z1_consul
...
```

in your stub.  Then, after generating the manifest with spiff, you can use some short scripts in something like Ruby to add your additional resource pools to the `resource_pools` section of the manifest.

```
manifest = YAML.load_file("manifest.yml")
pool_stub = YAML.load_file("pool-stub.yml")
new_pools = manifest["resource_pools"] + pool_stub["resource_pools"]
manifest["resource_pools"] = new_pools
puts YAML.dump(manifest)
```

The spiff templates are a convenience to get basic deployments going, but only satisfies a very slim set of use cases, e.g. it only supports two AZs.  They're not meant to support all use cases, and we want to avoid throwing every type of customizability into these templates as they are very complex and difficult to maintain.

v-boutes:
Ok I understand the philosophy behind the merging of templates, I close this pull request. Thank you @Amit-PivotalLabs for your answers and the time spent.

Amit-PivotalLabs:
Thanks @v-boutes! I appreciate your understanding.

poblin-orange:
@Amit-PivotalLabs agreed the templates are very complex, and difficult.

Any insight or pointers on cf-release future evolution (wrt to BOSH 2.0 constructs like cloud-config, bosh-links) which would ease this customization process  ?

Amit-PivotalLabs:
Sorry for the delayed response Pierre.  Yes, I plan to share some of the
work we've done and will be doing soon.

Cheers,
Amit

Amit-PivotalLabs:
Sorry, I should clarify.

I plan to share this information on the cf-dev and cf-bosh mailing lists:

<https://lists.cloudfoundry.org/archives/>
https://lists.cloudfoundry.org/archives/

Best,
Amit

PR #978, opened 2016-06-13
Loggregator etcd tls

nelsam:
- Adds properties for loggregator to work with etcd-tls
- Adds consul agents to templates that have metron agents
- Adds etcd TLS properties to template
- Changes `etcd.require_ssl` and `etcd.peer_require_ssl` to merge from stub with a default of false, rather than being forced to false

The first commit shouldn't have any changes that collide with changes from other teams.  The second commit is separate because it has changes related to more global properties, which may have conflicts with changes from other teams.

### DIFF
--- a/templates/cf.yml
+++ b/templates/cf.yml
@@ -1227,12 +1227,12 @@ meta:
     release: (( meta.loggregator_release_name ))
 
   clock_templates:
+  - name: consul_agent
+    release: (( meta.consul_release_name ))
   - name: cloud_controller_clock
     release: (( meta.capi_release_name ))
   - name: metron_agent
     release: (( meta.loggregator_release_name ))
-  - name: consul_agent
-    release: (( meta.consul_release_name ))
 
   consul_templates:
   - name: consul_agent
@@ -1251,22 +1251,22 @@ meta:
     release: (( meta.loggregator_release_name ))
 
   etcd_templates:
+  - name: consul_agent
+    release: (( meta.consul_release_name ))
   - name: etcd
     release: (( meta.etcd_release_name ))
   - name: etcd_metrics_server
     release: (( meta.etcd_release_name ))
   - name: metron_agent
     release: (( meta.loggregator_release_name ))
-  - name: consul_agent
-    release: (( meta.consul_release_name ))
 
   ha_proxy_templates:
+  - name: consul_agent
+    release: (( meta.consul_release_name ))
   - name: haproxy
     release: (( meta.cf_release_name ))
   - name: metron_agent
     release: (( meta.loggregator_release_name ))
-  - name: consul_agent
-    release: (( meta.consul_release_name ))
 
   hm9000_routes:
   - name: hm9000
@@ -1288,14 +1288,14 @@ meta:
     release: (( meta.cf_release_name ))
 
   loggregator_templates:
+  - name: consul_agent
+    release: (( meta.consul_release_name ))
   - name: doppler
     release: (( meta.loggregator_release_name ))
   - name: syslog_drain_binder
     release: (( meta.loggregator_release_name ))
   - name: metron_agent
     release: (( meta.loggregator_release_name ))
-  - name: consul_agent
-    release: (( meta.consul_release_name ))
 
   loggregator_trafficcontroller_routes:
   - name: doppler
@@ -1310,24 +1310,24 @@ meta:
     - (( "loggregator." .properties.system_domain ))
 
   loggregator_trafficcontroller_templates:
+  - name: consul_agent
+    release: (( meta.consul_release_name ))
   - name: loggregator_trafficcontroller
     release: (( meta.loggregator_release_name ))
   - name: metron_agent
     release: (( meta.loggregator_release_name ))
   - name: route_registrar
     release: (( meta.cf_release_name ))
-  - name: consul_agent
-    release: (( meta.consul_release_name ))
 
   nats_templates:
+  - name: consul_agent
+    release: (( meta.consul_release_name ))
   - name: nats
     release: (( meta.cf_release_name ))
   - name: nats_stream_forwarder
     release: (( meta.cf_release_name ))
   - name: metron_agent
     release: (( meta.loggregator_release_name ))
-  - name: consul_agent
-    release: (( meta.consul_release_name ))
 
   blobstore_routes:
   - name: blobstore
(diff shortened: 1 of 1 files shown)


templates/cf.yml
```diff
@@ -1210,6 +1231,8 @@ meta:
     release: (( meta.capi_release_name ))
   - name: metron_agent
     release: (( meta.loggregator_release_name ))
+  - name: consul_agent
```
fraenkel: consul_agent needs to be first everywhere

templates/cf.yml
```diff
@@ -618,6 +639,8 @@ properties:
     tls:
       ca_cert: ~
     etcd:
+      ca_cert: (( .properties.etcd.ca_cert ))
+      require_ssl: (( .properties.etcd.require_ssl ))
```
Amit-PivotalLabs: Don't you also need `cluster` here?

templates/cf.yml
```diff
@@ -631,6 +654,9 @@ properties:
     blacklisted_syslog_ranges: ~
     unmarshaller_count: (( merge || 5 ))
     port: (( merge || 4443 ))
+    etcd:
+      client_cert: ~
+      client_key: ~
```
Amit-PivotalLabs: I see this in 3 places.  For those 3 places, shouldn't it be `(( .properties.etcd.client_cert ))` instead of just `~`?

wfernandes: We actually didn't want to use the `etcd.client_cert` because we wanted separate etcd client certs for our components so that in the event of a leaked cert, we only need to roll the respective components. For example, if Doppler & Metron shared the same etcd client certs and if it were compromised, then we would have to roll the entire CF and Diego deployment.

Also there is another job called the `syslog_drain_binder`. We didn't add that in this template because it could get pulled in from the stub. So in the stub we would have an additional property block like below:

```
syslog_drain_binder:
  etcd:
    client_cert: foo
    client_key: bar
```

templates/cf.yml
wfernandes: The cluster configuration is directly under the etcd properties [line 582](https://github.com/cloudfoundry/cf-release/pull/978/files#diff-b9982126ee4438e20576a100670694d8R582). These properties are to provide our components (Doppler, TrafficController, MetronAgent, SyslogDrainBinder) with the etcd `ca_cert`, `require_ssl`, and `machines`

templates/cf.yml
```diff
@@ -97,6 +97,10 @@ jobs:
     properties:
       metron_agent:
         zone: z1
+      consul:
+        agent:
+          services:
+            etcd: {name: cf-etcd}
```
wfernandes: Just wanted to point this out, we named this `cf-etcd` in comparison to `other-etcd`. The certs should have common name of `cf-etcd.service.cf.internal`

templates/cf.yml
Amit-PivotalLabs: It sounds like some of the etcd properties you're copying over into the `loggregator.etcd` namespace, but stuff like `cluster` you're taking directly from the `etcd` namespace?

templates/cf.yml
Amit-PivotalLabs: I see.  So we'll need to generate a lot of different etcd client certs.  There's a question of what should be possible in the manifest, and what should be possible via the spiff templates.  We should be able to configure the client certs independently in the manifest.  Whether the spiff workflow should do that is a separate question.  The spiff templates have a confusing use cases, they're supposed to be both easy to get started with, but some people also want it to be hardened for production deployments too.  We err on the side of simplicity, because there's no way to guarantee spiff templates are production hardened.  Let's discuss whether it makes sense, for the purpose of these templates, to just copy around the same client certs.

templates/cf.yml
wfernandes: Yes. That is correct. The clients don't care about the cluster.

templates/cf.yml
wfernandes: That's fine with us. I just didn't want to make us dependent on the `etcd.client_cert` and `etcd.client_key` properties. We can create another etcd client cert and share it across the loggregator components if that is simpler.

shalako: @wfernandes if you intend to use `client_cert` and `client_key` namespaced to your client jobs, why does the PR add these properties to the top-level `etcd` namespace?

templates/cf.yml
Amit-PivotalLabs: I see.  The clients don't care about cluster, but they do need to know how to talk to the server, which is what "machines" is for.  Now this tricky.  In non-tls mode, the "machines" should be the list of IPs.  But in TLS mode, it should be the single DNS name of the secure etcd cluster, in this case "cf-etcd.service.cf.internal".  See e.g. https://github.com/cloudfoundry-incubator/diego-release/blob/develop/jobs/bbs/spec#L59-L61.  I'm not sure what the right thing to do here for spiff.

templates/cf.yml
Amit-PivotalLabs: @wfernandes Makes sense, want to allow different client certs for different jobs.  That's all well and good, but for the spiff templates, which are sorta geared to simplicity, they should probably just copy the same client cert around to all clients, rather than ask the new operator to produce N different client certs and keys.  So in spiff, I still stand by my original line comment here.

Amit-PivotalLabs:
Please make sure the unit tests pass on cf-release with these changes.  In its current state, this PR will fail.  The best thing to do is to run `./scripts/setup-git-hooks` from the `cf-release` directory on your workstation so that it runs unit tests whenever you try to commit.

Amit-PivotalLabs:
@nelsam @wfernandes Did you intend to continue working on this?  If not, you could simply wait until the Infrastructure team adds this stuff in anyways.  It might be a while, as we're working on the component to support zero-downtime switch to etcd in TLS mode.  In that case, we could close out this PR.

templates/cf.yml
wfernandes: @Amit-PivotalLabs Sounds good. We can use the same certs and tie them to the loggregator components that need them.

wfernandes:
@Amit-PivotalLabs We have stopped working on the etcd-tls epic unless/until we need to fix something. So no pressure from us. :)

Amit-PivotalLabs:
Closing this PR, we have this story: https://www.pivotaltracker.com/story/show/122070663.

Will link to this PR for context.

PR #979, opened 2016-06-13
BOSH-lite manifest template allows custom users

rosenhouse:
We would like to be able to add extra users to our bosh-lite deployment.

Other infrastructure templates support this.

### DIFF
--- a/templates/cf-infrastructure-bosh-lite.yml
+++ b/templates/cf-infrastructure-bosh-lite.yml
@@ -580,6 +580,7 @@ properties:
     scim:
       users:
       - admin|admin|scim.write,scim.read,openid,cloud_controller.admin,clients.read,clients.write,doppler.firehose,routing.router_groups.read
+      - <<: (( merge ))
 
   router:
     status:


templates/cf-infrastructure-bosh-lite.yml
```diff
@@ -580,6 +580,7 @@ properties:
     scim:
       users:
       - admin|admin|scim.write,scim.read,openid,cloud_controller.admin,clients.read,clients.write,doppler.firehose,routing.router_groups.read
+      - <<: (( merge ))
```
APShirley: Spiff adds an empty field when you don't add users resulting in this:
```yaml
scim:
  users:
  - admin|admin|scim.write,scim.read,openid,cloud_controller.admin,clients.read,clients.write,doppler.firehose,routing.router_groups.read
  - {}
```

This causes the following error when deploying on bosh-lite because the UAA template can't handle an empty entry in the scim.users field:
```
Error 100: Unable to render instance groups for deployment. Errors are:
   - Unable to render jobs for instance group 'uaa_z1'. Errors are:
     - Unable to render templates for job 'uaa'. Errors are:
       - Error filling in template 'uaa.yml.erb' (line 163: undefined method `+' for nil:NilClass)
```

APShirley:
Hi,

As there hasn't been any activity on this pull request for a week I'm going to close this for now. Feel free to open a new pull request if you have a fix for this issue.

Thanks,
Al

rosenhouse:
Yeah, I lost track of this, but your original point stands, I don't see how to make it work.

Thanks.

PR #995, opened 2016-06-28
Update cf-stub.yml for cf login failed with status code: 400

YuPengZTE:
root@boshcli:~/cf-release/scripts# cf login --skip-ssl-validation -a https://api.mybosh.com login -u admin -p c1oudc0w
API endpoint: https://api.mybosh.com
Authenticating...
Server error, status code: 400, error code: , message: 

in /var/vcap/jobs/uaa/config/uaa.yml require_https: true

PR #1000, opened 2016-07-04
Cf login failed with status code: 400

YuPengZTE:
When test for #995 ,with https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md,update and run manifest generation tests and other basic tests ./scripts/test, error is founded,so modify the value of expected.

Failures:

  1) Manifest Generation openstack behaves like generating manifests builds the correct manifest for openstack
     Failure/Error: expect(actual).to yaml_eq(expected)
       Mismatched values in ["properties", "login", "protocol"]:
          actual=https
        expected=http
     Shared Example Group: "generating manifests" called from ./spec/manifest_generation_spec.rb:29
     # ./spec/manifest_generation_spec.rb:16:in `block (3 levels) in <top (required)>'

Finished in 2.44 seconds (files took 0.23925 seconds to load)
4 examples, 1 failure

### DIFF
--- a/spec/fixtures/openstack/cf-manifest.yml
+++ b/spec/fixtures/openstack/cf-manifest.yml
@@ -1288,7 +1288,7 @@ properties:
     messages: null
     notifications:
       url: null
-    protocol: http
+    protocol: https
     restricted_ips_regex: null
     saml: null
     self_service_links_enabled: null
--- a/spec/fixtures/openstack/cf-stub.yml
+++ b/spec/fixtures/openstack/cf-stub.yml
@@ -77,7 +77,7 @@ properties:
   loggregator_endpoint:
     shared_secret: LOGGREGATOR_ENDPOINT_SHARED_SECRET
   login:
-    protocol: http
+    protocol: https
   nats:
     user: NATS_USER
     password: NATS_PASSWORD


Amit-PivotalLabs:
@sreetummidi @mbhave does this template change make sense?  Should it only be done for openstack?

mbhave:
@Amit-PivotalLabs as far as I can tell from the commits this change is only for fixtures. I don't see why the openstack fixtures should behave any differently because there is nothing related to the `login.protocol` in cf-infrastructure-openstack.yml.

YuPengZTE:
We test in openstack just,do not know the status in other IaaS.

Amit-PivotalLabs:
@mbhave these fixtures become the officially documented way to create stubs for CF.  Should the change made to the stub be what folks do by default?

mbhave:
@Amit-PivotalLabs The default for this property in the uaa-release spec file is `https`. https://github.com/cloudfoundry/uaa-release/blob/develop/jobs/uaa/spec#L92. I don't think it needs to be documented as part of the minimum number of properties that need to be set in order to get a working CF manifest.

YuPengZTE:
The CLA is ZTE -PaaS,I am a public member.

Amit-PivotalLabs:
The purpose of this PR is not clear.

PR #1017, opened 2016-07-25
allow `login.prompt` property to be overridden in template

wbean1:
* useful for when ldap uaa backend is being used and email
  address is not the user key
* see related https://github.com/cloudfoundry/cli/issues/810

### DIFF
--- a/templates/cf.yml
+++ b/templates/cf.yml
@@ -883,6 +883,7 @@ properties:
     invitations_enabled: ~
     spring_profiles: ~
     messages: ~
+    prompt: ~
 
     notifications:
       url: ~


Amit-PivotalLabs:
Hi @wbean1 

I suspect this will fail the unit tests.  Do you mind running them, and updating any fixture files that need to be updated to make the tests pass.  If you can squash the commits into one so that there is just one passing commit, that would be ideal.  Also, please close and re-open the PR once you believe you have signed and submitted the CLA.

For more info, please check out the [contributing guide](https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md).

Cheers,
Amit

PR #1018, opened 2016-07-26
CF template can optionally use postgres release

valeriap:
Update the cf.yml to allow overrides for the postgres job template.
This is because postgres has been moved out of cf-release.

Thanks,
@valeriap and @christianang


APShirley:
This looks good; it should obviously work just fine. Merging.

Signed-off-by: Jesse Alford <jalford@pivotal.io>

PR #1031, opened 2016-08-10
Add consul_agent to the api_z1 job in the example manifest stub

aduffeck:
We noticed that the `consul_agent` is missing from the `api_z1` job in the example stub at [https://docs.cloudfoundry.org/deploying/openstack/cf-stub.html](https://docs.cloudfoundry.org/deploying/openstack/cf-stub.html). Without it the api VM won't come up properly.

I also noticed that `java-buildpack` and `java-offline-buildpack` are missing and `cloud_controller_clock`  and `cloud_controller_worker` were added compared to the default values. Are those changes still supposed to be there?


ChunyiLyu:
Thanks for the contribution!

Chunyi & @APShirley

PR #1032, opened 2016-08-10
Make the update script work in a single pass.

davewalter:
The Release Engineering team has noticed that we have to run the update script multiple times in order to get a clean copy of a given version of cf-release. This seems to be a problem with cleaning up after updating the submodules of releases that are submoduled into this release.

To solve this problem, we reversed the order in which submodules are updated and cleaned, which allowed us to checkout any release and cleanse it with a single run of the script.


Amit-PivotalLabs:
Hey @davewalter, can you re-submit this PR against the develop branch please?

Thanks,
Amit

PR #1035, opened 2016-08-12
Add consul registration to trafficcontroller

apoydence:
This adds consul to loggregator trafficcontroller so that the TC can register itself with consul.  This shouldn't break anything on its own, but the route won't work properly until [185657bc13ab2ace36b370aa8505b39cf8e82fd8](https://github.com/cloudfoundry/loggregator/commit/185657bc13ab2ace36b370aa8505b39cf8e82fd8) is bumped for the health check script.

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml
+++ b/spec/fixtures/aws/cf-manifest.yml
@@ -775,8 +775,15 @@ jobs:
         - loggregator.SYSTEM_DOMAIN
     traffic_controller:
       zone: z1
+    consul:
+      agent:
+        services:
+          loggregator_trafficcontroller: {}
+
   resource_pool: small_z1
   templates:
+  - name: consul_agent
+    release: cf
   - name: loggregator_trafficcontroller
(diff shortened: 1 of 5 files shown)


Amit-PivotalLabs:
Thanks @apoydence 

Change looks good, but I suspect this will fail the unit tests.  You can take a look at the contributing docs [here](https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md).  Of relevance:

> 9. Set up git hooks by running `./scripts/setup-git-hooks` and commit your work. This will ensure that the same set of tests as `./scripts/test` run as a pre-commit hook before any commit.

Amit-PivotalLabs:
Also, once you get the tests to pass, please squash and resubmit PR as a single passing commit. Thanks!

apoydence:
Thanks @Amit-PivotalLabs, I updated the fixtures and the tests (scripts/test, smokes and cats) are all passing locally.

PR #1042, opened 2016-08-23
Add golang1.7 package and blobs

apoydence:
[#128422331]

Signed-off-by: Jason Keene <jkeene@pivotal.io>


Amit-PivotalLabs:
Hey @apoydence 

Sorry about the slow response.  Any reason you don't want to just commit this directly?  Or rather, since this commit appears to not include anything that actually *uses* this blob, do you want to simply include this change when you bump whatever else you're going to bump in cf-release that would actually consume the go 1.7 blob?  I presume that's loggregator, right?

iplay88keys:
We're closing this PR, we will just add it when we bump. Thanks @Amit-PivotalLabs

PR #1045, opened 2016-08-29
Adds UAA property to traffic controller

apoydence:
[#121083487]

Signed-off-by: Andrew Poydence <apoydence@pivotal.io>

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml
+++ b/spec/fixtures/aws/cf-manifest.yml
@@ -807,6 +807,9 @@ jobs:
         registration_interval: 20s
         uris:
         - loggregator.SYSTEM_DOMAIN
+    loggregator:
+      uaa:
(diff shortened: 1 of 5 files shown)


Amit-PivotalLabs:
Hi @apoydence 

This will break the unit tests.  Please see contributing docs: https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md.  In particular, you may want to:

> Set up git hooks by running ./scripts/setup-git-hooks and commit your work. This will ensure that the same set of tests as ./scripts/test run as a pre-commit hook before any commit.

Also, you should have access to commit changes directly to this repo.  No need to PR.

Finally, you might want to make your membership in the cloudfoundry GitHub org public.  That way @cfdreddbot will automatically detect that you are covered under CLA for any future pull requests.

PR #1048, opened 2016-09-05
Add script for deploying latest stable CF to BOSH Lite

jutkko:
Cache stemcell & cf release in WORKSPACE.

We are making our use of WORKSPACE explicit and believe this is better than relative paths. What do you think?


Amit-PivotalLabs:
Hi @jutkko 

This script doesn't seem to depend on the `cf-release` repository at all.  Seems like this script could live anywhere and still work.  If you find this script handy for your workflow, I encourage you to use it, but we don't want to maintain an additional script within this repo at this time.

Thanks,
Amit

PR #1049, opened 2016-09-06
Update loggregator properties

iplay88keys:
@Amit-PivotalLabs There are a few properties in Loggregator that have been deprecated for a while now. We believe that now is as good a time as any to remove them from the cf-release template. We wanted to make a PR to make sure that removing these properties won't break anything.

In the future, when we'd like to remove properties that we deprecated several versions ago, should we continue to make PRs, or just push directly to develop?

-Jeremy & Sam

### DIFF
--- a/templates/cf.yml
+++ b/templates/cf.yml
@@ -685,6 +685,7 @@ properties:
     blacklisted_syslog_ranges: ~
     unmarshaller_count: (( merge || 5 ))
     port: (( merge || 4443 ))
+    outgoing_port: (( merge || 8080 ))
     tls:
       server_cert: ~
       server_key: ~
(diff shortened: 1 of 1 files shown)


Amit-PivotalLabs:
Hey @iplay88keys 

Please commit this directly to develop.  Submitting as a PR won't really ensure that this doesn't break anything (aside from the Release Integration team doing some manual testing which you should have already done anyways).  If you want to be sure, review the CONTRIBUTING.md guide.  In particular, it describes installing git hooks so that whenever you commit to this repo it will run unit tests.  I strongly suspect this commit will cause unit tests to fail.

iplay88keys:
@Amit-PivotalLabs Yep, you're right.  It's odd that our computers don't have that hook installed already, but we'll set that up.

PR #1056, opened 2016-09-16
Break buildpack bosh releases into seperate repos

sesmith177:
The buildpacks team has updated our buildpack BOSH releases so that each releases lives in its own repo. This PR updates the submodules and symlinks so that cf-release uses the new repos. 

Tracker story: https://www.pivotaltracker.com/story/show/130042999


JesseTAlford:
This looks good to me, has passed CI I trust, and is a very exiting development.

Merging.

PR #1067, opened 2016-09-28
Scripts to generate CAs and certs for blobstore and uaa

thausler786:
When deploying CF to bosh-lite with the `skip_cert_verify` property set to false, certs are required for communication between some CF components.  The Cloud Controller in particular is not able to communicate with the uaa or the blobstore without certs.

These scripts create certificate authorities and use them to generate ssl certificates for the uaa and the blobstore that can be placed in the bosh director and used in the manifest to allow developers to use ssl when deploying bosh-lite with cert verification enabled.

https://www.pivotaltracker.com/story/show/129982147

Signed-off-by: Tim Hausler <thausler@pivotal.io>


staylor14:
Hi @thausler786,

Why two separate scripts, rather than one that takes an argument to differentiate the final file names?

thausler786:
@staylor14,

Thanks, that's a good idea, we'll change our scripts.  They were written this way following the pattern of the other scripts in cf-release, I'll drop a story in the cf-release backlog to clean up the other scripts as well.

jaresty:
Hey @thausler786. This looks great. One last thing: can we get rid of `generate-uaa-certs-old`? It's extra noise in the directory and we can always rely on git history if we want to see the original version of the script.

JesseTAlford:
Looks good to me!

PR #1080, opened 2016-10-20
Shorter cf oauth client RT validity

gberche-orange:
30 days is likely to be a too large default for most organizations (e.g. PWS sets 1 day)

In addition, a long refresh token validity (30 days) has the current side effect of growing the UAA db potentially up to its total capacity potentially creating outages, see https://github.com/cloudfoundry/uaa/issues/466: UAA stores all refresh tokens which have not expired or were invalidated.

### DIFF
--- a/example_manifests/minimal-aws.yml
+++ b/example_manifests/minimal-aws.yml
@@ -758,7 +758,7 @@ properties:
         authorized-grant-types: implicit,password,refresh_token
         autoapprove: true
         override: true
-        refresh-token-validity: 2592000
+        refresh-token-validity: 604800
         scope: cloud_controller.read,cloud_controller.write,openid,password.write,cloud_controller.admin,scim.read,scim.write,doppler.firehose,uaa.user,routing.router_groups.read
       cc-service-dashboards:
         authorities: clients.read,clients.write,clients.admin
--- a/spec/fixtures/aws/cf-manifest.yml
+++ b/spec/fixtures/aws/cf-manifest.yml
@@ -1526,7 +1526,7 @@ properties:
         authorities: uaa.none
         authorized-grant-types: password,refresh_token
         override: true
-        refresh-token-validity: 2592000
+        refresh-token-validity: 604800
         scope: cloud_controller.read,cloud_controller.write,openid,password.write,cloud_controller.admin,cloud_controller.admin_read_only,scim.read,scim.write,doppler.firehose,uaa.user,routing.router_groups.read,routing.router_groups.write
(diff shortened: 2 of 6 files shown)


Amit-PivotalLabs:
Hey @gberche-orange 

Thanks for this PR.  I personally don't totally understand those fields, but next to `600` you put the comment `1 hour`.  How does 600 translate to 1 hour?

This change will break the manifest generation unit tests.  Would you mind fixing the commit so that the unit tests pass, and squash that work into a single commit so the PR just consists of passing commits?  See steps 6 and 9 of the [General Workflow section of the Contributing doc](https://github.com/cloudfoundry/cf-release/blob/master/CONTRIBUTING.md#general-workflow) for more details on these tests.

Finally, could @sreetummidi and @plfx review this commit and sign off on whether the change is reasonable?

Thanks,
Amit

plfx:
The change is reasonable from the UAA's perspective. However, `600` seconds is ten minutes, rather than one hour. The refresh token time looks correct.

gberche-orange:
thanks @Amit-PivotalLabs and @plfx for reviewing this PR and sorry for having skipped the unit test step. 

I pushed a commit which fixes it, the invalid comment, along with the updated sample aws manifest.

dsabeti:
@gberche-orange Just to clarify, you'd like the default value to be 10 minutes instead of 1 hour? 

@plfx Is 10 minutes too low for a token validity interval?

Assuming 10 minutes is the intended change, and we don't think it's too low, we can merge this in. Otherwise, @gberche-orange, maybe we can change the value to 1 hour (or `3600` in terms of the manifest)?

sreetummidi:
@dsabeti Please hold off on merging this.
This affects all deployments including PWS. Let's talk when I am in office.

gberche-orange:
@dsabeti My request is only to reduce the refresh token duration (`refresh-token-validity` ) from 30 days to 7 days. I am not requesting to change current `access-token-validity` of 600 s (10 mins): I'm merely adding a comment to make is more human readeable in the template.

@sreetummidi I understand the PWS deployment would not be affected by this PR as the RT 30 days default seems currently overriden by PWS ops: RT tokens emmited on the `cf` client expire after 1 day.

dsabeti:
@sreetummidi Any updates on this?

sreetummidi:
@dsabeti I am good with this update.
Apologize for the delay !

dsabeti:
Great. @gberche-orange. The story is prioritized to be merged in. We'll try to get to it soon.

PR #1082, opened 2016-10-24
Add placeholders to make the etcd cluster run in secure TLS mode

djvdorp:
I wanted to update the [docs](https://docs.cloudfoundry.org/deploying/aws/cf-stub.html#stub) but it turns out that the cf-stub comes from `cf-release`, hence my Pull Request here. I tried to keep this change as concise and small as possible, and I am sorry in advance if I might have missed something.
# 

cf-release v241 release notes tell us:

[..]
This release introduces official support for running the etcd cluster
(shared by several components such as Routing API and the loggregator
subsystem, but not Diego which uses its own secure cluster) in secure
TLS mode. Upgrading an existing deployment with an insecure etcd cluster
to a secure one with minimal downtime is non-trivial. Instructions and
additional information for this procedure can be found here. If you are
using the manifest generation scripts included within the cf-release
repo to generate manifests, you're strongly recommended to upgrade to a
secure etcd cluster at this point. The instructions above assume you are
upgrading to a secure etcd cluster from a pre-v241 Cloud Foundry
deployment and will not apply as smoothly if you later attempt to
upgrade a post-v241 non-TLS etcd cluster to a TLS cluster within the
Cloud Foundry deployment.
[..]

See: https://github.com/cloudfoundry/cf-release/releases/tag/v241

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml
+++ b/spec/fixtures/aws/cf-manifest.yml
@@ -1295,8 +1295,8 @@ properties:
     blacklisted_syslog_ranges: null
     debug: false
     etcd:
-      client_cert: ""
-      client_key: ""
+      client_cert: ETCD_CLIENT_CERT
+      client_key: ETCD_CLIENT_KEY
     maxRetainedLogMessages: 100
     message_drain_buffer_size: null
     port: 4443
@@ -1314,9 +1314,9 @@ properties:
     enabled: true
   etcd:
     advertise_urls_dns_suffix: cf-etcd.service.cf.internal
-    ca_cert: ""
-    client_cert: ""
-    client_key: ""
+    ca_cert: ETCD_CA_CERT
+    client_cert: ETCD_CLIENT_CERT
+    client_key: ETCD_CLIENT_KEY
     cluster:
     - instances: 2
       name: etcd_z1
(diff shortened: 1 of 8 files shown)


dsabeti:
Hey @djvdorp Thanks for this PR. I think it makes sense to get these changes into the docs for the stub. Before we can accept the change I have a few requests:

1. Could you add the same fields to the other stubs (vsphere, openstack, bosh-lite)?
2. The cf-release unit tests are currently failing as a result of your change. It seems that they're failing to parse the yaml, and from what I can tell it seems like a whitespace issue. Once you fix that, the tests will still likely fail because the changes also need to be reflected in `spec/fixtures/<infrastructure>/cf-manifest.yml` -- the test uses the stub fixture to generate a manifest, and then asserts that result matches the expected manifest. You can run the tests with `scripts/test` or `bundle exec rspec spec`, and hopefully the output should guide you in making changes.

djvdorp:
Hi @dsabeti Thank you for your comments, they have been processed (the same changes for the other stubs + fixed whitespace + fixed expectations to that the tests pass again). I hope this change will help many others in the future!

staylor14:
Hi @djvdorp!  It appears that you've made these changes against a branch of master.  Can you please make a pull request against develop instead?  

Thanks very much!

djvdorp:
@staylor14 will do that now!

PR #1084, opened 2016-10-27
Add dotnet-core-buildpack BOSH release

sesmith177:
- Add src/dotnet-core-buildpack-release submodule
- Add job, package symlinks
- Update spec fixtures
- Update templates
- Upload dotnet-core-buildpack v1.0.3 blob

[#131997427]

Signed-off-by: Jacques Chester jchester@pivotal.io

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -61,3 +61,6 @@
 [submodule "src/java-buildpack-release"]
 	path = src/java-buildpack-release
 	url = https://github.com/cloudfoundry/java-buildpack-release
+[submodule "src/dotnet-core-buildpack-release"]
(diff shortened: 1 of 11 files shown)


dsabeti:
@sesmith177 It looks like this PR got rebased (instead of merged) by mistake, but you can see that the commit is on develop: https://github.com/cloudfoundry/cf-release/commit/9f456dd29c330be0cb621f50b17a458086939fca

@jaresty && @staylor14, does this sound right?

If we've confirmed that all the necessary changes are in develop, can @sesmith177 close the PR?

jaresty:
@dsabeti That sounds correct.  Thanks!

Issue #1089, opened 2016-10-31
Cannot configure `artifacts_directory` from the manifest for smoke tests


henrytk:
## Expected behaviour

When setting `properties.smoke_tests.artifacts_directory` in the manifest it is expected to be [templated into the smoke tests configuration file](https://github.com/cloudfoundry/cf-release/blob/v245/jobs/smoke-tests/templates/config.json.erb#L20-L37), with some logic which only adds it if it has been set.

## Actual behaviour

It does not.

## Impact

Implementors who want to configure an artifacts directory either have to use their own template or create workarounds, such as manipulating the configuration file after it has been templated.

## Suggested solution

Add logic to the template which checks if it has been set in the properties and add it if it has. I would not expect it to [hard code the value as is done for the acceptance tests](https://github.com/cloudfoundry/cf-release/blob/v245/jobs/acceptance-tests/templates/config.json.erb#L24) for the following reasons:

* this forces it to be on by default
* smoke tests can be run with much higher frequency, therefore generating a lot of output.

dsabeti:
Hi @henrytk. You're correct that `artifacts_directory` is a field in the configuration as defined by the smoke test source code, but you may notice that it has never been part of [the errand's configuration definition](https://github.com/cloudfoundry/cf-release/blob/master/jobs/smoke-tests/spec#L15-L56). In general, an internal component (in this case, smoke-tests) consuming a certain field does not necessitate the ability to configure that field via the bosh manifests -- in fact, in some cases we purposefully do this to simplify the (already bloated) bosh manifests.

Having said that, it seems like a totally reasonable thing to add to the interface of smoke-tests errand. Would you be willing to submit a PR?

dsabeti:
Hi @henrytk. I'm going to close this issue since we've prioritized your PR: https://github.com/cloudfoundry/cf-release/pull/1090

PR #1090, opened 2016-11-02
Allow configuration of artifacts directory for smoke tests

henrytk:
## Why

This pull request has been raised regarding https://github.com/cloudfoundry/cf-release/issues/1089

It is useful to be able to ship the CF_TRACE output to temporary storage, so it can be used to debug failed tests. This pull request has been raised to allow the configuration of the artifacts directory from the manifest, which will enable the logging more easily.

## What

Add some logic to check if the `smoke_tests.artifacts_directory` property is set and add it to the smoke test configuration. This makes it possible to turn on logging of CF_TRACE output from the manifest.

Once the template is rendered you will be left with some fairly hideous whitespace (a linefeed before and after the artifacts directory key). I did not implement trimming of this because the Bosh repository shows inconsistent support for the `trim_mode` parameters in ERB[1].

The artifacts directory is hard coded for the acceptance tests job[2]. Other than a mild inconsistency there is no reason to change this behaviour.

[1] http://ruby-doc.org/stdlib-2.3.1/libdoc/erb/rdoc/ERB.html#method-c-new
[2] https://github.com/cloudfoundry/cf-release/blob/v245/jobs/acceptance-tests/templates/config.json.erb#L24

### DIFF
--- a/jobs/smoke-tests/spec
+++ b/jobs/smoke-tests/spec
@@ -49,6 +49,8 @@ properties:
   smoke_tests.enable_windows_tests:
     default: false
     description: Toggles a portion of the suite that exercises Windows platform support
+  smoke_tests.artifacts_directory:
+    description: The directory in which to store test CF_TRACE output.
   smoke_tests.backend:
     default: ''
     description: Defines the backend to be used. ('dea', 'diego', '' (default))
--- a/jobs/smoke-tests/templates/config.json.erb
+++ b/jobs/smoke-tests/templates/config.json.erb
@@ -18,6 +18,9 @@
   end
 %>
 {
+<% if_p("smoke_tests.artifacts_directory") do |artifacts_directory| %>
+  "artifacts_directory"  : "<%= artifacts_directory %>",
+<% end %>
   "suite_name"           : "<%= properties.smoke_tests.suite_name %>",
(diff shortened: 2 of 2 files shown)


jaresty:
Hi @henrytk,

I'm working on this pull request and would like to use the new property you've added.  Do you mind  giving a step-by-step example with concrete values of how you will use it once it is merged?

Thanks!

Josh

henrytk:
Hi @jaresty 

Our use case is a little strange because we don't run the smoke tests as an errand, we use the template to run the smoke tests in a Concourse pipeline.

We will add `properties.smoke_tests.artifacts_directory` to [our manifest](https://github.com/alphagov/paas-cf/blob/staging-0.0.321/manifests/cf-manifest/manifest/900-cf-tests.yml#L24), then we will use the template in [our Concourse pipeline](https://github.com/alphagov/paas-cf/blob/staging-0.0.321/concourse/tasks/smoke-tests-config.yml#L36-L40) to generate the configuration.

After the files are generated we use [a Concourse task to ship them to an S3 bucket](https://github.com/alphagov/paas-cf/blob/staging-0.0.321/concourse/tasks/upload-test-artifacts.yml#L16-L29).

Hope that helps.

PR #1104, opened 2016-11-16
Added spruce version of core yml and openstack derivations

stevewallecs:
I added files for the spruce version of some manifest files as opposed to updating the current versions. This update includes the core cf.yml file, which is used by other IaaS derivations. I didn't have time to do all the IaaS files, so I just added the ones that I did and included "-spruce" in the name.

### DIFF
--- a/spec/fixtures/openstack/cf-stub-spruce.yml
+++ b/spec/fixtures/openstack/cf-stub-spruce.yml
@@ -1,0 +1,206 @@
+# The following line helps maintain current documentation at http://docs.cloudfoundry.org.
+# code_snippet cf-stub-openstack start
+---
+director_uuid: DIRECTOR_UUID
+
+meta:
+  environment: ENVIRONMENT
+
+  floating_static_ips:
+  - 173.1.1.1
+
+networks:
+  - name: floating
+    type: vip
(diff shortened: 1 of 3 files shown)


dsabeti:
Hey @stevewallecs, thanks for this PR.

The Release Integration team (the team responsible for cf-release and manifest generation) is moving away from these scripts entirely in favor of [cf-deployment](https://github.com/cloudfoundry/cf-deployment). In the meantime, we're only planning to support the tooling that we've already had for a while and are testing in automated builds. As such, I'm reluctant to merge in a change that a) introduces a previously unsupported tool (spruce) to cf-release, and b) that our team doesn't have the bandwidth to add testing for.

On the other hand, I'm sure that a lot of people in the CF community would be excited to see more tooling around manifest generation until we officially bless cf-deployment for public consumption. I'd suggest sending an email to cf-dev@lists.cloudfoundry.org to let people know that you have a fork of cf-release that offers better support for spruce.

Let me know if this works for you, or if you have any other questions regarding manifest generation in cf-release.

David Sabeti
CF Release Integration PM

stevewallecs:
Thanks for the response David. I see your points and look forward to cf-deployment!

PR #1105, opened 2016-11-21
Allow router resource pools to receive extra properties

luan:
- This enables using a bbl'd up environment for deploying from cf-release using its manifest generation
- Also incidentally allow extra cloud_properties to be added to router resource pools

Signed-off-by: Aakash Shah <ashah@pivotal.io>


JesseTAlford:
Thanks, @luan and @aashah! This will be very helpful for people using `cf-release` with `bbl`.

PR #1107, opened 2016-11-23
Pr/required uaa config in manifest generator

plfx:
Generate key and self-signed cert for now-required manifest properties of the UAA. Other properties look like they are already being created. The generated cert can be overridden with stubs


plfx:
We were not pulled down. I will make a new PR.

JesseTAlford:
Cool, I deleted the corresponding tracker story.

PR #1124, opened 2016-12-09
Add "monitor fail" to haproxy health check

philippthun:
In case no servers (i.e. routers) are available, haproxy is still reporting "HTTP 200 - Service ready". When using this health endpoint in an IAAS load balancer it would make more sense to report an error (HTTP 503) instead, so that the load balancer does not forward any requests to this instance.

Signed-off-by: Amin Chawki <amin.chawki@sap.com>


ChunyiLyu:
@philippthun Thanks for your PR! We do ask contributors to open PR only on develop branch. Would you mind closing this PR, and reopen it again against develop?

Thanks,
Chunyi

PR #1127, opened 2016-12-16
Update loggregator certs and allow for metron cert/key in properties

jtarchie:
[#136223987]

Signed-off-by: JT Archie <jtarchie@gmail.com>


JesseTAlford:
This looks good to me. Merging.

In the future it would be helpful if y'all could detail any testing you may have already done - running manifest generation, provisioning a bosh-lite, etc.

PR #1131, opened 2016-12-21
Further highlight that the admin user password should be replaced

aemengo:
The current deployment works by making an admin user with a password
of 'PASSWORD'

### DIFF
--- a/example_manifests/minimal-aws.yml
+++ b/example_manifests/minimal-aws.yml
@@ -415,7 +415,7 @@ jobs:
         userids_enabled: true
         users:
           - name: admin
-            password: PASSWORD
+            password: REPLACE_WITH_PASSWORD
             groups:
               - scim.write
               - scim.read


JesseTAlford:
We should totally do this. It's consistent with how every other cred place-holder is handled.

I'd merge it right now, but we have some automation that needs to be updated to target the new name. We'll do it soon though.

JesseTAlford:
@aemengo, I extended this throughout the manifest. We weren't just using PASSWORD in one spot; it was all over the place. I changed them _all_ to REPLACE_WITH_PASSWORD. (We use string substitution in our CI, and it makes it easier if they're all the same. Also, they _should_ all be replaced. Probably not with the same password, but hey, minimal effort for minimal security.)

Thanks again for seeing this.

aemengo:
@JesseTAlford Awesome, glad to see it happen.

PR #1134, opened 2016-12-30
Move clock_global into clock_z{1,2} to enable HA mode

luan:
-- **DO NOT MERGE UNTIL https://www.pivotaltracker.com/story/show/136621261 IS RELEASED** --

This requires https://www.pivotaltracker.com/story/show/136621261 to be
released and merged into cf-release before it can be merged.

The clock job will hold onto a lock in the CCDB to ensure only one clock
is scheduling jobs at a time, which enables deploying more than one
instance of it.

### DIFF
--- a/spec/fixtures/aws/cf-manifest.yml
+++ b/spec/fixtures/aws/cf-manifest.yml
@@ -485,7 +485,7 @@ jobs:
 - default_networks:
   - name: cf1
   instances: 1
-  name: clock_global
+  name: clock_z1
   networks:
   - name: cf1
   persistent_disk: 0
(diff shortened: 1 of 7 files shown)


cppforlife:
can it be colocated with CCs then instead of having separate machines?

aashah:
@cppforlife Sure, it's still okay to colocate the CC jobs (api, worker, clock) together.

aashah:
The story has been released if this could be merged please.

Thanks,
Aakash & Luan

aashah:
@cppforlife I think I misunderstood the question.

We probably could remove the clock job entirely in favor of colocating them with the API instances, but realistic deploys have many, many more api VMs than we'd need clocks. The approach for having the clock be HA includes maintaining an open transaction to the DB, so that would also increase if colocated.

SocalNick:
@dsabeti bump - we intended this to go out with CF-251, can we merge it?

dsabeti:
Sure, the PR looks fine to merge. One question though: if clock is HA, does it make sense to co-locate it with the `api_*` jobs? Would like to avoid increasing the footprint if possible.

dsabeti:
After talking with @utako and @luan, they brought up the valid point that it doesn't make sense to scale the clocks with the api (imagine a deployment with dozens of cloud controllers). I'll merge the PR.

PR #1076, opened 2016-10-11
added Google Compute Engine template

poligraph:
Created Google Compute Engine template and updated generate script.

PR #1145, opened 2017-01-27
Add softlayer example manifest

petergtz:
This PR is a proposal to provide an example manifest for Softlayer in the same way an example exists for AWS.

A stale version of the example manifest is living in https://github.com/cloudfoundry/bosh-softlayer-cpi, but we believe that cf-release is a better place to host this example to avoid becoming stale again. The idea is to delete stale version in the other location.

The example manifest presented here was tested in a real Softlayer environment and works with CF 251.

We realize that the efforts to migrate to cf-deployment and the use of the new BOSH CLI will make this addition possibly obsolete, but we think this could still add value meanwhile in the same way that giving people specific instructions to deploy into AWS adds value.

We'd be happy to get feedback on this proposal.

Thanks,
Peter

/cc @smoser-ibm @suhlig


dsabeti:
Hey @petergtz. Thanks for this PR. We've had a few requests like this in the last couple of months, where people want to contribute manifests and templates for a new IaaS. Here's an example of what I told previous requesters: https://github.com/cloudfoundry/cf-release/pull/1076#issuecomment-258566675

In short, I'm hesitant to publish a manifest or template whose functionality our team can't guarantee because we don't test it in CI, and cf-release is going to be deprecated soon anyway so we're focusing on [cf-deployment](https://github.com/cloudfoundry/cf-deployment). On the other hand, I don't want this work to go to waste either.

There are a few options available to us here:
1. You could put this softlayer manifest in your own repo and announce on the cf-dev mailing list (cf-dev@lists.cloudfoundry.org) and the Cloud Foundry slack. If anybody asks us questions about deploying to softlayer, we could point them to your repo.
2. Take a look at [cf-deployment](https://github.com/cloudfoundry/cf-deployment). This repo contains The Future:tm: for deploying CF. You could help to write an operations file that enables users to deploy to softlayer, and we could work on accepting that contribution.

Let me know whether one, or both, of these solutions would work for you.

petergtz:
Hi @dsabeti. What you say makes sense. Familiarizing and migrating to cf-deployment is already on our roadmap.

For the time being, we'll host this minimal example for Softlayer in the [bosh-softlayer-cpi](https://github.com/cloudfoundry/bosh-softlayer-cpi).

Until cf-release went from deprecation to deletion, would you be willing to add a section to the [README](https://github.com/cloudfoundry/cf-release/blob/master/example_manifests/README.md) that provides links to unsupported minimal example CF manifests? Something along the lines of:

> For similar instructions with other cloud providers see this [list of unsupported minimal example manifests](link-to-a-section-in-the-of-this-README).

That could be added to the end of the Introduction section. It would make it clear that these are unsupported examples, but at the same time make them visible to anyone. I could imagine that teams building other CPIs would be interested too.

/cc @smoser-ibm @suhlig

dsabeti:
@petergtz Sounds perfectly reasonable. Here's the change on `develop`: https://github.com/cloudfoundry/cf-release/commit/5dea85c341f9aa56beb787e0e18a4913ee5398e7

I linked to the minimal doc in the bosh-softlayer-cpi repo, but once you host the manifest there I'd be happy to update the link to point to the manifest itself.

If everything looks good, go ahead and close the PR.

PR #1146, opened 2017-01-31
CAPI jobs consume nil nats links

wendorf:
- cloud_controller_ng
- cloud_controller_worker
- cloud_controller_job

This is to allow the job specs to optionally consume the link. Without
this change, the jobs don't know which of the multiple nats jobs to
consume the link from.


zankich:
Thanks!

PR #1152, opened 2017-02-15
Consume ssh_proxy_link on ha_proxy config if available

luan:
This depends on https://github.com/cloudfoundry/diego-release/pull/273 being merged first.

Now we disable consul templating in favor of bosh links and more advanced HAProxy config.


cppforlife:
consider giving it more specific type so that implicit linking connects up components automatically without addiotional manifest configuration. 

cc @dsabeti

dsabeti:
Hey @luan. 

I just realized that there will be two jobs (access_z1 and access_z2) that both provide this link in a standard Diego deployment. I may be wrong, but I'm pretty sure that this PR will fail most deployments because BOSH won't know which of the two links to consume. I think you'll have to nil out the link in cf-release manifest (like we do with consul and nats).

In which case, why have this change, if you have to deploy it such that link cannot be consumed?

dsabeti:
@luan Bump.

dsabeti:
Hey @luan, I'm going to close this PR if it looks like you don't need it.

luan:
Hey @dsabeti. Yeah you're right, but that's only true if you're using a cf-release generated manifest.
I know of consumers of `ha_proxy` that use the job from cf-release in a cf-deployment type manifest (only one `access` VM).

We do have to nil out the link in cf-release manifest unfortunately for that to work.

dsabeti:
Hey @luan. Thanks for the context. It's good point, so I just prioritized the story to merge this.

PR #1169, opened 2017-03-15
Fix typos in tls certificate generation doc

petergtz:
I used this guide to generate our certificates and found some typos along the way which this PR fixes.


dsabeti:
Good catch! Thanks for submitting the PR.

PR #1170, opened 2017-03-15
Add errand to run smoke-tests on windows

mdelillo:
This errand can be run on a windows machine. It depends on https://github.com/cloudfoundry/cf-smoke-tests/pull/25.

@mdelillo @pnikonowicz

### DIFF
--- a/config/blobs.yml
+++ b/config/blobs.yml
@@ -1,4 +1,6 @@
----
+cli/cf-cli_6.23.1_winx64.zip:
+  size: 6759874
+  sha: 22c09fbaaaed4ed305896de28a3c1bf9b346ff93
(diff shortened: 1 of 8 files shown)


staylor14:
Hi @mdelillo!

Thanks for the pull request!  One note regarding adding new blobs: if you let Bosh do the upload, the changes will get automagically added to config/blobs.yml.  The change you committed lacked an object ID, which cause our unit tests to fail: https://release-integration.ci.cf-app.com/teams/main/pipelines/cf-release/jobs/unit-tests/builds/21

We pushed the fix, so we're all set.

Thanks again!
--Steve && @ChunyiLyu

mdelillo:
That's good to know, thanks for fixing it!

PR #1175, opened 2017-03-29
Don't consume etcd link

nmahoney-pivotal:
Once a new etcd-release is cut with links, we don't want to consume it within cf-release.

Thanks,
Christian and Nick

### DIFF
--- a/templates/cf.yml
+++ b/templates/cf.yml
@@ -1433,6 +1433,7 @@ meta:
     consumes: {consul: nil}
   - name: etcd
     release: (( meta.etcd_release_name ))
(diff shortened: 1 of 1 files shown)


ChunyiLyu:
Hi @nmahoney-pivotal && @christianang, thanks for the PR! The unit tests `scripts/test` has failed after merging the PR, because our fixture manifests could no longer match generated manifests. We have pushed the fix to those fixture files, so everything is good now :) 

In the future, we would appreciate if you could run the unit test first.

The Relint Team

christianang:
Ah missed that, will remember in the future. Thanks.

PR #1196, opened 2017-05-02
Changed default ciphers

lcacciagioni:
Updated default ciphers to match the defaults of the haproxy-boshrelease ([here](https://github.com/cloudfoundry-incubator/haproxy-boshrelease/blob/master/jobs/haproxy/spec#L52)). This ciphers give a baseline for clients like: Firefox 1, Chrome 1, IE 7, Opera 5, Safari 1, Windows XP IE8, Android 2.3, Java 7.


dsabeti:
Hi @lcacciagioni. Thanks for the PR. I understand the motivation to support different versions of different browsers, but the existing set of ciphers were chosen for security purposes. Opening it up would lead to some security vulnerabilities, and we can't allow that as a default configuration for haproxy.

One thing I'll point out is that the values provided in the `spec` file are only _defaults_ -- you can always override them in your manifest. Just know that you'd potentially be opening yourself up to security issues if you did so.

I'll leave this PR open for a little while in case you have some follow up questions or other thoughts you want to bring up, but as is, it's unlikely that we'll merge this in.

lcacciagioni:
@dsabeti I know that this is for security as it is, but it seems like a little restrictive more if you want to interact from browsers, anyway I think that you are right. Drop it if you fell it.

PR #1213, opened 2017-06-13
Allow variables section to be specified in stub

anEXPer:
If operators wish to use the variable related features in the bosh CLI, this should allow them to specify a variables section in their stub and have it appear in the final manifest.

### DIFF
--- a/templates/generic-manifest-mask.yml
+++ b/templates/generic-manifest-mask.yml
@@ -15,3 +15,5 @@ compilation: (( merge ))
 update: (( merge ))
 
 resource_pools: (( merge ))
+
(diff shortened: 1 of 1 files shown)


anEXPer:
Reopening so the bot can see I am legit.

dsabeti:
I _really_ don't want to add a major feature to the cf-release manifest generation scripts. We're all moving forward to cf-deployment, and I'd rather focus engineering work / support load on that. I'm going to close this PR, but if anyone feels strongly that `variables` support in cf-release is valuable to them, I'd like to hear feedback.

PR #1251, opened 2017-09-26
Refactor certificates generation scripts

paolostivanin:
- Recreating the CA by default will lead to a downtime if the "multi-step" deployment approach is not followed. This PR introduces a new paramter (`--recreate-ca`) that, if passed to
the scripts, will generate/recreate also the CA.
By default now all the scripts will skip the CA generation and will use the existing one instead.

- fix shellcheck issues

- use `/usr/bin/env bash` as shebang, so MAC users have less issues

- create a script called `generate-cf-certs`, similar to what diego-release does


scripts/generate-loggregator-certs
```diff
@@ -1,65 +1,72 @@
...
-certstrap --depot-path ${depot_path} request-cert --passphrase '' --common-name trafficcontroller
-certstrap --depot-path ${depot_path} sign trafficcontroller --CA loggregator-ca
+certstrap --depot-path "${depot_path}" request-cert --passphrase '' --common-name trafficcontroller
+certstrap --depot-path "${depot_path}" sign trafficcontroller --CA "$ca_name"
 
 # Traffic Controller<->CAPI mutual TLS certificate
-certstrap --depot-path ${depot_path} request-cert --passphrase '' --common-name cc_trafficcontroller
-certstrap --depot-path ${depot_path} sign cc_trafficcontroller --CA ${bbs_ca_name}
+certstrap --depot-path "${depot_path}" request-cert --passphrase '' --common-name cc_trafficcontroller
+certstrap --depot-path "${depot_path}" sign cc_trafficcontroller --CA "${bbs_ca_name}"
```
dsabeti: When I tried to run `./scripts/generate-cf-certs .`, it failed on this line consistently. It looks like `generate-loggregator-certs` is failing to find the Diego BBS CA (`cf-diego-ca`):
```
stat /Users/davidanthonysabeti/workspace/cf-release/loggregator-certs/cf-diego-ca.crt: no such file or directory
```

(I think it expects the argument to `--CA` to be the _path_ to the CA file. Before this refactoring, the user was required to copy that CA cert into the `loggregator-certs` directory. Now that we're scripting it, `generate-cf-certs` will probably need to find a way to copy that file to the correct location.)

paolostivanin: My bad, I didn't consider the `.` case :) I was testing it with `$PWD`. Unfortunately, certstrap accepts only abs path and not rel path, this is way it's failing if you pass `.`
I will fix it, good catch 👍  

What the script actually does is to symlink the diego CA and CAkey to the current dir, so certstrap can sign the cert. This is way the copying is no longer needed.

scripts/generate-cf-certs
```diff
@@ -0,0 +1,18 @@
+#!/usr/bin/env bash
+
+set -e
+
+# Install certstrap
+go get -v github.com/square/certstrap
+
+pushd "$(dirname "$0")/.." > /dev/null
+    creds_folder="$1" && shift
```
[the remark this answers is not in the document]
paolostivanin: Ok 👍

on-tls-certificates.md
```diff
@@ -72,12 +72,23 @@ used by the Cloud Controller's interactions with other components.
...
+  ...
+```
+
 ### Start by generating certs for the small "cliques"
 
 #### Consul
 Generate the certs and keys:
 ```
-./scripts/generate-consul-certs
+./scripts/generate-consul-certs <ROOT_DEPOT_PATH> [--recreate-ca]
```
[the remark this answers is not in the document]
paolostivanin: You're right, I'll fix it :)

scripts/generate-loggregator-certs
```diff
@@ -2,71 +2,63 @@
...
-  echo -e "\e[1m\e[33m[INFO] Using existing loggregator CA, recreating only client/server certificates\e[0m"
+if [[ -z "$3" || "$3" != "--no-bbs-ca" ]]; then
+  bbs_ca_cert_path="$(find "$ROOT_DEPOT_PATH" -name 'cf-diego-ca.crt')"
+  bbs_ca_key_path="$(find "$ROOT_DEPOT_PATH" -name 'cf-diego-ca.key')"
+  if [[ -z "$bbs_ca_cert_path" || -z "$bbs_ca_key_path" ]]; then
+    echo -e "\e[1m\e[33m[ERROR] Please execute '$SCRIPTS_FOLDER/generate-cf-diego-certs' before running this script.\e[0m"
+    exit 1
+  fi
+  bbs_ca_cert_path="$(realpath "$bbs_ca_cert_path")"
+  bbs_ca_key_path="$(realpath "$bbs_ca_key_path")"
```
dsabeti: `realpath` isn't present on mac systems. Can we use a generic function that works on multiple platforms for this?

dsabeti:
Hey @paolostivanin. Sorry for all this delayed back and forth. I just took another stab at using the script, and it looks like certstrap is having a hard time following the symlink:
```
Get CA certificate error: stat /Users/davidanthonysabeti/workspace/cf-release/statsd-injector-certs/loggregator-ca.crt: no such file or directory
```

It looks like certstrap doesn't like symbolic links, so you may need to update the script to copy them instead of symlink them.

dsabeti:
Hey @paolostivanin, since cf-release has been deprecated, I'm going to go ahead and close this issue.

PR #1257, opened 2017-10-20
Not default to debug2 logging level in CC

mikexuu:
debug2 logging level is too verbose and logs database queries. This could expose secrets.

### DIFF
--- a/templates/cf.yml
+++ b/templates/cf.yml
@@ -744,8 +744,8 @@ properties:
     internal_api_user: "internal_user"
     internal_api_password: (( bulk_api_password ))
 
-    logging_level: debug
-    db_logging_level: debug
+    logging_level: info
+    db_logging_level: off
 
     staging_upload_user: (( merge ))
     staging_upload_password: (( merge ))
--- a/templates/cf.yml
+++ b/templates/cf.yml
@@ -744,8 +744,8 @@ properties:
     internal_api_user: "internal_user"
     internal_api_password: (( bulk_api_password ))
 
-    logging_level: debug2
-    db_logging_level: debug2
+    logging_level: debug
+    db_logging_level: debug
 
     staging_upload_user: (( merge ))
     staging_upload_password: (( merge ))


dsabeti:
Hey @mikexuu. This is a pretty straightforward change, but I'm worried that the impact could be quite large if operators have been depending on this behavior. Can you currently override those values in the stubs? If not, maybe the better change would be to make these values configuration, but leave the default at their current levels.

mikexuu:
@dsabeti Yeah making these properties configurable through stubs would be good. However, CAPI told us that setting db_logging_level to debug2 enables query logging, which could expose db salts and encrypted creds. 
Maybe the ideal thing to do here is to make it configurable but merge on the CC defaults? (off for db logging and info for normal logging).

dsabeti:
@mikexuu Ok, if you update the PR to the levels you requests, but allowed for spiff to override them -- something like `logging_level: (( merge || "info" ))` -- we could merge this, and just let people know in release notes that they can keep the original configuration by updating their stubs.

mikexuu:
@dsabeti I just spoke with the CAPI team. CC's current bosh templating doesn't let you specify `off`, but in their dev machines where they do have it off & have syslog-forwarder enabled, apparently you will start getting seg-faults. 
Maybe this whole PR is premature right now, we can revisit this when CAPI pushes out a new fix. 👍

ljfranklin:
@mikexuu the next capi-release CAPI 1.45.0 will turn DB logging off by default unless the user sets `cc.log_db_queries: true`. I think this PR can be closed as the default values in the next release will not log db queries.

mikexuu:
@ljfranklin Ah I see, interesting. 
I wonder if there's value in having _some_ DB logging for security auditing purposes (where we can determine if any user has performed CVE exploit X). Is there a long-term solution for CAPI to continue DB logging but without sensitive data like encrypted values + salts?

ljfranklin:
@mikexuu good question. I think our initial thought was that many users would consider most of the non-encrypted fields like org names to be "secret" so the focus was on disabling all DB logs by default. CC'ing @zrob to see if a feature where we selectively redact encrypted fields could be a further enhancement.

As far this PR is concerned, I think we can leave both logging properties at `debug2` and if the user really wants to see the DB logs they will also have to turn on `cc.log_db_queries`.

mikexuu:
Sounds good, thanks for looking into this @ljfranklin !

